PR #21, opened 2022-12-01
ToBytes opcode

vezenovm:
This specifies the `ToBytes` opcode in the GadgetCall match statement. This was updated in both the static_lib and wasm packages. 

There was also an additional line in the `common` package for the serialiser. I used the `_` for the match statement, but perhaps this should be changed to specifying the exact OPCODE being serialised so we don't accidentally reach that match branch for OPCODEs that should be implemented in the serialiser.


TomAFrench:
Closing as the `toBits` opcode has been included in `noir-lang/noir` (as `toRadix`)

PR #25, opened 2022-12-06
Update bb commit reference to renamed aztec-2.0

vezenovm:
This branch is just here to back port the Noir backend as we update to the new barretenberg repository. The old repo was renamed and we need this new commit reference in order for the main Noir repo to pass its CI build.


kevaundray:
Closing as we no longer reference the old backend

PR #33, opened 2023-01-02
Fix CRS file size check and incorporate it into the Wasm backend

ax0:
The CRS file size check is only present in the static lib and always results in a download whenever `nargo prove` or `nargo verify` is run. This PR changes the check to ensure that it passes if the following `g2_data` assignment makes sense. Moreover, the check is moved to the CRS logic in `common`, thus incorporating it into the Wasm backend, and `common`'s CRS code is brought into scope in `barretenberg_static_lib` to avoid duplicate code.


ax0:
Yup. Should I just move the tests over to lib.rs, or would you prefer I put them somewhere else?

vezenovm:
I think they are fine where they are for now. We can always move them if needed. I'm going to merge this in a moment as I need this fix as well.

PR #36, opened 2023-01-09
Use published acvm v0.1.0

phated:
This replaces the `acvm` dependency from the noir repo and with the `acvm` from its standalone repo.

This dependency currently can't be downloaded on a fresh install because the `7460dda929130589bccfd6619f9409834b72a135` commit was part of https://github.com/noir-lang/noir/pull/551 but that branch was deleted—so a fresh clone of `noir-lang/noir` doesn't include the commit.

@vezenovm suggested that we point to the https://github.com/noir-lang/acvm repository now that it has been split from the noir repo; however, there has been *a lot* of changes to the project. I've pinned the project to the same commit as the final, shared commit: https://github.com/noir-lang/noir/commit/c15c9dabc7d402b54ea290d2b6e0572f513865bd

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,26 +5,44 @@ version = 3
 [[package]]
 name = "acir"
 version = "0.1.0"
-source = "git+https://github.com/noir-lang/noir?rev=7460dda929130589bccfd6619f9409834b72a135#7460dda929130589bccfd6619f9409834b72a135"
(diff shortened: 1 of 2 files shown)


phated:
Chatted with @kevaundray, who said that the published 0.1.0 version should be a compatible version. So I changed this PR to use the published version because it is cleaner.

kevaundray (approved):

PR #37, opened 2023-01-11
Update barretenberg_wrapper to use branch

phated:
This updates the `barretenberg_wrapper` crate from noir-lang/aztec-connect to the latest commit. This is necessary to build successfully on Mac with homebrew in a non-default location.

We may want to wait to update this until we land https://github.com/noir-lang/aztec-connect/pull/8 so we have the Clang changes for Linux ARM

### DIFF
--- a/barretenberg_static_lib/Cargo.lock
+++ b/barretenberg_static_lib/Cargo.lock
@@ -5,26 +5,44 @@ version = 3
 [[package]]
 name = "acir"
 version = "0.1.0"
-source = "git+https://github.com/noir-lang/noir?rev=7460dda929130589bccfd6619f9409834b72a135#7460dda929130589bccfd6619f9409834b72a135"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b5b510cf711ccaedb4c56172a52745e9f1cbc2d88731430ceccd2d6eb5b2f6cd"
 dependencies = [
+ "acir_field",
  "flate2",
  "indexmap",
- "noir_field",
  "rmp-serde",
+ "serde",
+]
+
+[[package]]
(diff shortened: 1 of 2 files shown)


phated:
Now that https://github.com/noir-lang/aztec-connect/pull/8 and https://github.com/noir-lang/aztec-connect/pull/14 have been merged, I've updated this to the latest commit to ensure we have all the fixes we need.

jfecher (commented):
I'd be in favor of replacing the rev entirely with `branch = "kw/noir-dsl"` to keep this updated. Thoughts?

phated:
> I'd be in favor of replacing the rev entirely with branch = "kw/noir-dsl" to keep this updated. Thoughts?

I'd be in favor of that, but I believe @kevaundray told me that he'd prefer refs throughout the projects. Maybe I misunderstood.

kevaundray:
We can change this to branch, for something like nargo referencing the aztec_backend, refs would be better so that changes can be made to the aztec_backend without breaking the package manager (Its possible to use a develop branch for ongoing changes to backend)

phated:
> so that changes can be made to the aztec_backend without breaking the package manager

How does cargo handle transitive dependencies? If you use a branch here, but a ref to `aztec_backend`, would the underlying `barretenberg_wrapper` on a `cargo update`?

phated:
I've updated this to use the branch.

kevaundray:
> > so that changes can be made to the aztec_backend without breaking the package manager
> 
> How does cargo handle transitive dependencies? If you use a branch here, but a ref to `aztec_backend`, would the underlying `barretenberg_wrapper` change on a `cargo update`?

Yep that was my understanding

PR #38, opened 2023-01-12
Fix clippy warnings

TomAFrench:
This just removes unused imports and uses slices instead of vecs where they're not needed.


jfecher (approved):
👍

phated (approved):
Thanks!

PR #39, opened 2023-01-15
Migrate to acvm 0.3 API

kevaundray:
This is currently blocked as we are using it as a git dependency, instead of a published acvm version


barretenberg_static_lib/src/composer.rs
```diff
@@ -133,7 +133,7 @@ impl StandardComposer {
         if let Some(pi) = &public_inputs {
             let mut proof_with_pi = Vec::new();
             for assignment in pi.0.iter() {
-                proof_with_pi.extend(&assignment.to_bytes());
+                proof_with_pi.extend(&assignment.to_be_bytes());
```
vezenovm, line 136: This change did not break anything? If we were formatting pub inputs as little endian before and must switch to big endian I am a little confused how we were able to verify proofs successfully before.

vezenovm, line 136: This change did not break anything? If we were formatting pub inputs as little endian before and must switch to big endian I am a little confused how we were able to verify proofs successfully before.

barretenberg_static_lib/src/composer.rs
```diff
@@ -227,26 +227,10 @@ mod test {
             public_inputs: None,
             result: false,
         };
-        // This should fail as we specified that we do not have any public inputs
```
vezenovm, line 230: Why did we get rid of these tests?

barretenberg_static_lib/src/composer.rs
```diff
@@ -478,75 +464,74 @@ mod test {
 
     //     test_circuit(constraint_system, vec![case_1]);
     // }
```
vezenovm, line 466: Can we delete this schnorr test in the lines above as well or update the values and uncomment the code? I just noticed this comment above the test: 
```
    // TODO: an added asset statement was added, making this test fail
    // TODO: we can regenerate a signature using the new code
```
which is one of the errors I ran into when I updated the backend a couple months ago and has been fixed now. However, this test does a full prove and verify rather than simply interop with barretenberg like in the `schnorr.rs` file

vezenovm: Because we updated the pedersen tests to work I'm leaning towards updating this schnorr test and uncommenting it

barretenberg_static_lib/src/lib.rs
```diff
@@ -8,12 +8,11 @@ mod pippenger;
 pub mod scalar_mul;
 pub mod schnorr;
 use common::acvm::FieldElement;
-use std::convert::TryInto;
 
 pub struct Barretenberg;
 
 pub fn field_to_array(f: &FieldElement) -> [u8; 32] {
-    let v = f.to_bytes();
+    let v = f.to_be_bytes();
```
vezenovm, line 15: Once again, just hoping this does not break anything else. Can we make sure this backend upgrade is also tested using all the tests under `nargo/tests/test_data`?

barretenberg_static_lib/src/composer.rs
kevaundray: Good question -- so the new API is mostly cosmetic changes. So `to_bytes` was always using big endian. Its just that now, we make it explicit. 

So the only thing I did in acvm was change:

```rust
fn to_bytes(&self) -> Vec<u8>
```

to

```rust
fn to_be_bytes(&self) -> Vec<u8>
```

barretenberg_static_lib/src/composer.rs
kevaundray: The proving system's behaviour has changed -- If you specify that there will be 2 public inputs and you do not pass two public inputs, then it will panic or abort if it is using wasm

barretenberg_static_lib/src/lib.rs
kevaundray: Replied above -- in Noir there is migrate to 0.3 PR open that runs the tests

barretenberg_static_lib/src/composer.rs
vezenovm: Schnorr test looks to be passing now so resolving this convo

vezenovm (approved):
Looks good to me. Tests passing and most upstream changes seem to be related to naming. It also looks to me like any other concerns with the migration can be resolved using the new separate acvm repo.

PR #42, opened 2023-01-18
Remove console_error_panic_hook from barretenberg_static_lib

phated:
This dependency seems to be leftover from before the `aztec_backend_wasm` was split out into its own package. Should be good to remove from here.


kevaundray (approved):
LGTM

PR #43, opened 2023-01-18
Migrate CI to use reusable workflows

phated:
This migrates our GitHub Actions to use reusable actions (available at https://github.com/noir-lang/.github/tree/main/.github/workflows).

I also took the opportunity to replace the `actions-rs/toolchain` action with https://github.com/dtolnay/rust-toolchain which seems to be an up-to-date action with similar behavior (and [David Tolnay](https://github.com/dtolnay) is a highly respected Rust dev). This means that we no longer have deprecation warnings in every run.

Note that `cargo test` in the reusable workflows were updated to always run with `--test-threads=1` in https://github.com/noir-lang/.github/commit/f8b71908808ca081bd58b4c058d3bd06888b201c because the tests in this project were failing when running in parallel and we want to keep things uniform.

You'll also note that I removed `.github/pkg` which I think was a mistaken check-in, but let me know if I'm wrong.


kevaundray (approved):

vezenovm (approved):

vezenovm:
Once CI finishes looks like we are good to merge

phated:
Thanks. I'm actually unhappy with that it runs the clang10 build per entry in the matrix so I'm going to refactor a bit. Sorry for the delay.

vezenovm:
> Thanks. I'm actually unhappy with that it runs the clang10 build per entry in the matrix so I'm going to refactor a bit. Sorry for the delay.

Sounds good. No worries! Ping me when I should take another look.

vezenovm:
> You'll also note that I removed `.github/pkg` which I think was a mistaken check-in, but let me know if I'm wrong.

Should this instead be in [aztec_backend_wasm](https://github.com/noir-lang/aztec_backend/tree/master/aztec_backend_wasm)?

kevaundray:
> > You'll also note that I removed `.github/pkg` which I think was a mistaken check-in, but let me know if I'm wrong.
> 
> Should this instead be in [aztec_backend_wasm](https://github.com/noir-lang/aztec_backend/tree/master/aztec_backend_wasm)? @kevaundray

yes, though its fine to delete it from here

vezenovm:
> > > You'll also note that I removed `.github/pkg` which I think was a mistaken check-in, but let me know if I'm wrong.
> > 
> > 
> > Should this instead be in [aztec_backend_wasm](https://github.com/noir-lang/aztec_backend/tree/master/aztec_backend_wasm)? @kevaundray
> 
> yes, though its fine to delete it from here

Sound like we are good to merge then @phated

PR #44, opened 2023-01-19
use acvm version which introduces public outputs [DO NOT MERGE]

kevaundray:
This is not mergeable since we are using an unpublished version of acvm


TomAFrench:
Closing as noir-lang/noir is now running on a version of ACVM which uses public outputs

PR #46, opened 2023-01-24
feat: Leverage rustls when using downloader crate

phated:
The openssl crate makes it difficult to cross-compile our dependencies, so this swaps the TLS implementation to [rustls](https://github.com/rustls/rustls).

This is a dependency that doesn't require openssl to exist on a system, and helps with cross-compilation. I found out that they don't have complete support (e.g. RISC-V is unsupported) but they seem to support all the platforms we really care about.


phated:
Need a review from @kobyhallx since @kevaundray deferred to him about this change

kobyhallx (approved):
LGTM 👍

PR #48, opened 2023-01-27
update bb.wasm file for hash_to_field fix

vezenovm:
This is an update to resolve this issue on the main Noir repo: https://github.com/noir-lang/noir/issues/662

While migrating to `noir-lang/aztec-connect/barretenberg` I fixed the `hash_to_field` interop in barretenberg. However, due to us currently having a manual build process for the `barretenberg.wasm` file I missed that I needed to rebuild the wasm file for the aztec backend. The wasm aztec backend is what is currently used for binary releases so that is why the error still existed in the issue linked. I tested this update locally and was able to use `hash_to_field` with the wasm aztec backend and not just the static lib.


phated (approved):
Thanks

PR #49, opened 2023-02-01
Remove timing printlns

jfecher:
This removes the printouts in the form `Total Proving time (Rust + Static Lib) : 40469750ns ~ 0seconds` which are largely unnecessary and notably clog the output when using the new `nargo test` command.

Users interested in proving and verification time can test it themselves via `time`, `hyperfine`, or other more featureful command line utilities.


vezenovm:
@kevaundray I am for removing these as they clutter the output for `nargo test`

kevaundray:
> @kevaundray I am for removing these as they clutter the output for `nargo test`

Sounds good -- I think if we want to do debugging, then they can be added back manually or we can add a compile flag in the future

kevaundray (approved):

PR #56, opened 2023-02-17
feat!: receive `public_inputs` when verifying as a `BTreeMap` [DO NOT MERGE]

TomAFrench:
Following on from conversation on Slack, this PR updates the backend to make use of the new interface as defined in https://github.com/noir-lang/acvm/pull/96

The major benefit being that it's now semantically explicit that we're passing witness assignments in here and so duplicate values are impossible.

Something I want to flag up is that I haven't updated `serialise_public_inputs` in `aztec_backend_wasm`. https://github.com/noir-lang/aztec_backend/blob/cff757dca7971161e4bd25e7a744d910c37c22be/aztec_backend_wasm/src/lib.rs#L94

The reasoning for this is that we also take a vector for the initial witness in `compute_witnesses` so it seems that we want to avoid maps here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,8 +5,7 @@ version = 3
 [[package]]
 name = "acir"
 version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


TomAFrench:
Closing in favour of #58

PR #59, opened 2023-02-22
fix: rename gates to opcodes

TomAFrench:
We're referring to the old "gates" field on `Circuit` rather than the new "opcodes" field. This should be picked up be type checking so we should also investigate why this doesn't error.


kevaundray (approved):

kevaundray:
aztec_backend_wasm is not included in the compilation targets. If you check Cargo.toml, it gets excluded. It was needed for the typescript code, but its not clear to me if typescript is using newer versions of this

PR #61, opened 2023-02-23
fix(aztec_backend_wasm): Sort the witness_map after solving

vezenovm:
In the TS packages, there were reported issues of basic programs returning `false` on proof verification. 
For example:
```
fn main(x: u64, y: u64) {
  constrain x < y;
}
``` 
was failing verification. However, this circuit would pass partial witness generation. This told us the witness mapping was correct, but rather the witness array being passed to barretenberg in the TS packages differed from the one being passed in nargo. 

The witness array we compute in `aztec_backend_wasm` is missing the step of adding witnesses in the correct order. Some circuits would still verify correctly such as:
```
fn main(x: u64, y: u64) {
  constrain x != y;
}
``` 
But this would only be because the circuit's witnesses were consecutive, so any missing values in the witness map came at the end and were `0` anyway. 

I added code to check whether the solved witness map has all the witness values, if it does not we fill the value with a zero. This led to the first snippet above verifying true in typescript. 

There were other test cases with merkle_membership that were passing partial witness generation but failing verification. With this addition they are all now passing verification.

### DIFF
--- a/aztec_backend_wasm/src/lib.rs
+++ b/aztec_backend_wasm/src/lib.rs
@@ -45,15 +45,22 @@ pub fn compute_witnesses(
         Err(opcode) => panic!("solver came across an error with opcode {}", opcode),
     };
 
-    // Serialise the witness in a way that the C++ codebase can deserialise
-    let assignments = crate::barretenberg_structures::Assignments::from_vec(
-        witness_map
-            .into_iter()
-            .map(|(_, field_val)| field_val)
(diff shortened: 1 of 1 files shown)


vezenovm:
Closing as @kobyhallx has incorporated these changes into this branch (https://github.com/noir-lang/aztec_backend/tree/kh-wasm-bindgen). The changes have been tested and previously failing proofs are now passing verification

PR #66, opened 2023-02-27
chore: simplifications and deduplications of shared code

TomAFrench:
We never pass `None` as the public inputs, it's always `Some` empty `Assignments` when we have no public inputs, I've then removed this extra layer.

I've also pulled all the code related to prepending and stripping public inputs to/from proofs into `common` so all methods can reuse it.

### DIFF
--- a/common/src/proof.rs
+++ b/common/src/proof.rs
@@ -1,6 +1,6 @@
 use crate::barretenberg_structures::Assignments;
 
-pub fn remove_public_inputs(num_pub_inputs: usize, proof: Vec<u8>) -> Vec<u8> {
+pub fn remove_public_inputs(num_pub_inputs: usize, proof: &[u8]) -> Vec<u8> {
     // This is only for public inputs and for Barretenberg.
     // Barretenberg only used bn254, so each element is 32 bytes.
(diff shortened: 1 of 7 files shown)


common/src/proof.rs
```diff
@@ -0,0 +1,22 @@
+use crate::barretenberg_structures::Assignments;
+
+pub fn remove_public_inputs(num_pub_inputs: usize, proof: Vec<u8>) -> Vec<u8> {
+    // This is only for public inputs and for Barretenberg.
+    // Barretenberg only used bn254, so each element is 32 bytes.
+    // To remove the public inputs, we need to remove (num_pub_inputs * 32) bytes
+    let num_bytes_to_remove = 32 * num_pub_inputs;
+    proof[num_bytes_to_remove..].to_vec()
```
jfecher, line 8: Since we don't need to own the Vec here, we could save one `.to_vec()` call of the callers by taking in `proof: &[u8]` instead.

Either that or we could take advantage of the fact we own proof and do `proof.drain(0 .. num_bytes_to_remove);` to avoid allocating a Vec. Perhaps this option is less safe though considering the unsafe code that surrounds the creation of the proof Vec in the callers.

jfecher (approved):

PR #68, opened 2023-02-27
chore: cleanup unused dependencies

TomAFrench:
We had a number of unused dependencies on various crates so I've removed any that we don't need and created dev-dependencies from those only used in tests.


jfecher (approved):
This is a lot of unused crates 👍

PR #71, opened 2023-02-28
chore: remove unused `barretenberg_js` crate

TomAFrench:
This PR removes the unused `barretenberg_js` crate from the workspace.


TomAFrench:
> Do we not use barretenberg_js for wasm anymore? What is the new backend for wasm?

Nope, we don't have any references to this crate outside of itself:https://github.com/search?q=org%3Anoir-lang%20barretenberg_js&type=code

Are you thinking of [barretenberg[dot]js](https://github.com/noir-lang/aztec-connect/tree/kw/noir-dsl/barretenberg.js)?

phated:
> Ah sorry, there's the barretenberg.wasm blob in barretenberg_wasm. I'm not sure on the source of that.

`barretenberg.wasm` is generated by cmake in the aztec-connect project and copy-pasted into this project. That will be handled via nix once we are using upstream bb.

TomAFrench:
Going to merge this as it's uncontroversial and its dependency has already been deleted.

PR #73, opened 2023-03-01
chore: mark `common` private and fix spellcheck issues

TomAFrench:
I've marked `common` to be a private crate so that we can change its interface without the change being breaking. I've taken advantage of this to do some of the spell check corrections which were previously breaking.


jfecher (approved):
Looks good. Unfortunate there are still public functions that misspell `barrtenberg`

PR #75, opened 2023-03-06
chore: factor out function to flatten witness map

TomAFrench:
We've currently got several different implementations of flattening a witness map into a vector in various crates (including one incorrect one in `aztec_backend_wasm`).

This PR pulls out this logic into a function in `common` which can be reused.

I've avoided making this a method on `Assignments` as this implementation is specific to proving whereas we use `Assignments` for both proving and verifying.


jfecher (approved):
Thank you

PR #77, opened 2023-03-06
chore: bump to ACVM 0.6.0

TomAFrench:
This PR updates us to use the new version of the ACVM crate. The only change which is relevant is the addition of the `Block` opcode.

@guipublic It seems to me that we can just handle this similarly to directives when serializing for BB. Let me know if we should handle this differently.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "acir"
-version = "0.5.0"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


guipublic:
> @guipublic It seems to me that we can just handle this similarly to directives when serializing for BB. Let me know if we should handle this differently.

No, we need to provide the block opcode to BB, there should be already be something for this.
May be we can skip them for now as you did, because noir assumes the opcode is not supported by the backend and properly handle them in another PR?

guipublic (approved):
That's fine for me, as I said in a comment, the backend will need to handle the block opcode but it can/should be done in another PR.

PR #79, opened 2023-03-08
feat(wasm): Reenable wasm backend

kobyhallx:
Need to re-enable wasm backend to publish backend


aztec_backend_wasm/src/lib.rs
```diff
@@ -21,7 +23,7 @@ pub fn compute_witnesses(
 ) -> ComputedWitness {
     console_error_panic_hook::set_once();
 
-    let circuit: Circuit = circuit.into_serde().unwrap();
+    let circuit: Circuit = JsValueSerdeExt::into_serde(&circuit).unwrap();
```
jfecher, line 26: Why the change from method to function syntax? Is there another `into_serde` on the circuit type?

aztec_backend_wasm/src/lib.rs
```diff
@@ -42,32 +44,46 @@ pub fn compute_witnesses(
     // Now use the partial witness generator to fill in the rest of the witnesses
     // which are possible
 
+    use barretenberg_wasm::Plonk;
     let plonk = Plonk;
+    let num_witnesses = circuit.num_vars();
```
jfecher, line 49: Nitpick: I don't think we need the `use` when the only line that uses Plonk is line 48

aztec_backend_wasm/src/lib.rs
TomAFrench: This is matching a change we performed in https://github.com/noir-lang/noir/pull/605

tl;dr: there's a circular dependency in `wasm-bindgen` which is resolved by doing this.

Cargo.toml
```diff
@@ -1,11 +1,14 @@
...
-members = ["common", "barretenberg_wasm", "barretenberg_static_lib"]
 
-exclude = ["aztec_backend_wasm"]
+members = ["common", "barretenberg_wasm", "barretenberg_static_lib", "aztec_backend_wasm"]
+# barretenberg_js will be phased out in the future
 
-[workspace.package]
+# exclude = ["aztec_backend_wasm"]
+
+exclude = ["barretenberg_js"]
```
TomAFrench, line 9: References to `barretenberg_js` are stale.

Cargo.toml
```diff
@@ -1,11 +1,14 @@
 [workspace]
 
-members = ["common", "barretenberg_wasm", "barretenberg_static_lib"]
 
-exclude = ["aztec_backend_wasm"]
+members = ["common", "barretenberg_wasm", "barretenberg_static_lib", "aztec_backend_wasm"]
+# barretenberg_js will be phased out in the future
 
-[workspace.package]
+# exclude = ["aztec_backend_wasm"]
```
TomAFrench, line 7: Can delete this commented line as well.

aztec_backend_wasm/src/lib.rs
```diff
@@ -5,6 +5,8 @@ pub use barretenberg_wasm::Barretenberg;
 
 use wasm_bindgen::prelude::*;
 
+use gloo_utils::format::JsValueSerdeExt;
```
TomAFrench, line 8: Can we add `gloo` to cspell?

aztec_backend_wasm/src/lib.rs
```diff
@@ -42,32 +44,46 @@ pub fn compute_witnesses(
...
+        let value = match witness_map.get(&Witness(i)) {
+            Some(value) => *value,
+            None => FieldElement::zero(),
+        };
+
+        sorted_witness.push(value);
+    }
 
-    assignments.to_bytes()
+    sorted_witness.to_bytes()
```
TomAFrench, line 69: This change needs to be reverted. This logic has been factored out into `proof::flatten_witness_map`

barretenberg_wasm/src/lib.rs
```diff
@@ -15,7 +15,6 @@ pub mod scalar_mul;
 pub mod schnorr;
 
 pub use common::crs;
-use std::cell::Cell;
```
TomAFrench, line 18: We make use of this on L80. Shouldn't we make that a qualified usage of `Cell` if we're removing this import?

TomAFrench (changes requested):

phated:
I believe this was superseded by #94

PR #80, opened 2023-03-09
chore: use qualified name for `std::cell::Cell`

TomAFrench:
Addresses https://github.com/noir-lang/aztec_backend/pull/79#discussion_r1129927419


jfecher (commented):
Why is this a separate PR instead of fixing #79 directly?

phated:
We closed #79 so I think we should close this too 🤔

PR #82, opened 2023-03-13
chore!: Remove create_proof and verify functions

phated:
This removes the `create_proof` and `verify` functions and reimplements the `acvm_interop` functions with @vezenovm's pk/vk functions.

I believe we also need to deprecate the acvm_interop stuff and eventually remove it from ACVM and here.


TomAFrench (commented):
We have a similar implementation in `barretenberg_wasm`. Do we not want to remove this at the same time?

TomAFrench (approved):
LGTM

vezenovm (approved):

PR #84, opened 2023-03-15
feat!: Depend upon upstream barretenberg & switch to UltraPlonk

vezenovm:
This updates aztec_backend to using upstream barretenberg, and switches from using TurboPlonk to UltraPlonk.

To streamline the installation of upstream bb, we are using Nix. This has lead us to replace the CI with nix commands, but if we want to add pure cargo commands, we can add more jobs that install bb and its deps globally.


phated:
The tests should all be passing now. The last thing should just be to finish the solidity templating and 🤞 this will be good to go.

barretenberg_wasm/src/lib.rs
```diff
@@ -235,32 +178,84 @@ impl Barretenberg {
...
+
+    // Print it!
+    println!("{string}");
+}
+
+// TODO: This should actually be creating random data, right?
+// Probably duplicate https://github.com/wasmerio/wasmer/blob/2.3.0/lib/wasi/src/syscalls/mod.rs#L2537
+fn random_get(_: i32, _: i32) -> i32 {
+    0_i32
+}
```
phated, line 211: @kevaundray what should be the priority on fixing this? Through some logging, it seems that random_get is actually getting called quite a bit

barretenberg_static_lib/Cargo.toml
```diff
@@ -11,7 +11,7 @@ rust-version.workspace = true
 [dependencies]
 
 common = { path = "../common" }
-barretenberg_wrapper = { git = "https://github.com/noir-lang/aztec-connect", branch = "kw/noir-dsl" }
+barretenberg-sys = { git = 'https://github.com/noir-lang/barretenberg-sys.git', branch = "mv/ultra" }
```
phated, line 14: TODO: Update to published version

.github/workflows/rust.yml
```diff
@@ -8,14 +8,32 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  check_n_test:
-    name: cargo check & test
-    uses: noir-lang/.github/.github/workflows/rust-test.yml@main
+  nix:
```
phated, line 11: Are we fine with **only** Nix testing or do we also want a globally installed barretenberg test? Like https://github.com/noir-lang/barretenberg-sys/blob/279be62243b8c51159b75960d536760d69d83990/.github/workflows/test.yml#L45-L95

barretenberg_wasm/src/lib.rs
kevaundray: I think we can merge without it with a tracking issue if this is going to take too much time to fix

barretenberg_wasm/src/composer.rs
```diff
@@ -346,7 +362,7 @@ mod test {
         // This fails because the constraint system requires public inputs,
         // but none are supplied in public_inputs. So the verifier will not
         // supply anything.
-        let case_1 = WitnessResult {
+        let _case_1 = WitnessResult {
```
kevaundray, line 365: can we remove the prepending underscore?

barretenberg_wasm/src/lib.rs
```diff
@@ -125,108 +131,45 @@ impl Default for Barretenberg {
...
 
     let custom_imports = imports! {
         "env" => {
-            "logstr" => log_env,
+            "logstr" => Function::new_native_with_env(
+                &store,
+                Env {
+                    memory: memory.clone(),
+                },
+                logstr,
```
kevaundray, line 151: Nice

barretenberg_wasm/src/composer.rs
phated: Not without removing the tests. These are broken at barretenberg, which panics.

barretenberg_wasm/src/lib.rs
TomAFrench:
```suggestion
// TODO: Implement randomization function https://github.com/noir-lang/aztec_backend/issues/101
```

Now we have the issue, let's add a link to it.

flake.nix
```diff
@@ -0,0 +1,151 @@
...
+      };
+
+      # We use `include_str!` macro to embed the solidity verifier template so we need to create a special
+      # source filter to include .sol files in addition to usual rust/cargo source files
+      solidityFilter = path: _type: builtins.match ".*sol$" path != null;
+      sourceFilter = path: type:
+        (solidityFilter path type) || (craneLib.filterCargoSources path type);
+
+      commonArgs = {
+        # TODO: Rename repository
```
kevaundray, line 66: Are these TODOs still relevant?

kevaundray (commented):
This LGTM -- I have left some comments but nothing blocking

barretenberg_wasm/src/lib.rs
```diff
@@ -235,32 +178,84 @@ impl Barretenberg {
...
+
+    // Convert the subslice to a `&str`.
+    let string = std::str::from_utf8(&str_vec).unwrap();
+
+    // Print it!
+    println!("{string}");
+}
+
+// TODO: This should actually be creating random data, right?
+// Probably duplicate https://github.com/wasmerio/wasmer/blob/2.3.0/lib/wasi/src/syscalls/mod.rs#L2537
```
TomAFrench, line 208:
```suggestion
// TODO: Implement randomization function https://github.com/noir-lang/aztec_backend/issues/101
```

vezenovm:
This looks good to me. Can't approve my own request so commenting here

barretenberg_wasm/src/lib.rs
phated: I might just finish it before this lands 🙃

.github/workflows/rust.yml
phated: Sounds like everyone is fine with this change. This will diverge from the reusable workflows until we get nix throughout all repos.

kobyhallx (approved):
LGTM 👍

PR #85, opened 2023-03-15
chore: Duplicate static_lib tests into wasm to verify same functionality

phated:
This adds additional testing to `barretenberg_wasm`. While working on #84, @vezenovm has to link everything into Noir to run tests to find problems in the wasm. This adds a base layer of tests so we hopefully catch some things before needing to link into Noir.


vezenovm (approved):
As these tests are mostly the exact same it would be nice to generalize them between the static_lib and wasm crates in some way. However, I'm not sure if it is as simple to do as stating it, so this is sufficient for now imo

PR #86, opened 2023-03-15
feat: allow overriding transcript location with BARRETENBERG_TRANSCRIPT env var

phated:
This change allows us to override the transcript location using the `BARRETENBERG_TRANSCRIPT` environment variable. This is necessary because Nix builds everything in a read-only sandbox where the download would fail. We can instead specify it as a dependency of building in nix and make the path available through this env var.

### DIFF
--- a/common/src/crs.rs
+++ b/common/src/crs.rs
@@ -1,3 +1,5 @@
+use std::{env, path::PathBuf};
+
 #[allow(clippy::upper_case_acronyms)]
 pub struct CRS {
     pub g1_data: Vec<u8>,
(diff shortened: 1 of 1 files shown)


common/src/crs.rs
```diff
@@ -9,12 +11,17 @@ const G1_START: usize = 28;
...
+fn transcript_location() -> PathBuf {
+    match env::var("BARRETENBERG_TRANSCRIPT") {
+        Ok(dir) => PathBuf::from(dir),
+        Err(_) => {
+            let mut transcript_dir = dirs::home_dir().unwrap();
+            transcript_dir.push(std::path::Path::new("noir_cache"));
+            transcript_dir.push(std::path::Path::new("ignition"));
+            transcript_dir.push(std::path::Path::new("transcript00.dat"));
+            transcript_dir
+        }
```
TomAFrench, line 23:
```suggestion
        Err(_) => dirs::home_dir()
            .unwrap()
            .join("noir_cache")
            .join("ignition")
            .join("transcript00"),
```

I've noticed that we tend to use this mutable "create many paths and push" pattern over "chain joins of strings" in quite a few places. I find this to be more readable so is there a reason why not to use it?

phated: Seems fine. I didn't know the API existed but it does the same thing internally, so let's do it.

TomAFrench (approved):

PR #89, opened 2023-03-23
feat(wasm): Oracle support

joss-aztec:
Wasm bindings for solving the intermediate witness with a callback for resolving oracle opcodes.


common/src/serializer.rs
```diff
@@ -341,16 +341,19 @@ pub fn serialize_circuit(circuit: &Circuit) -> ConstraintSystem {
             Opcode::Directive(_) => {
                 // Directives are only needed by the pwg
             }
-            Opcode::Block(_, _) => {
+            Opcode::Block(_) => {
                 // TODO: implement serialization of blocks to match BB's interface
             }
+            Opcode::ROM(_) | Opcode::RAM(_) | Opcode::Oracle(_) => {
+                todo!("Unimplemented")
```
joss-aztec, line 348: @vezenovm, I'm guessing Block/ROM/RAM will be addressed separately as part of your ultraplonk work.

I guess it should be possible to serialise oracle opcodes now though. Reckon you'll have time to give me a had with that?

joss-aztec: Thinking about it, I reckon oracles don't actually add any constraints. I'll stub it alongside the `Directive` case.

vezenovm: > I'm guessing Block/ROM/RAM will be addressed separately as part of your ultraplonk work.

Yes you can leave these unimplemented for now

> Thinking about it, I reckon oracles don't actually add any constraints. I'll stub it alongside the `Directive` case.

You are correct here as well, they should follow what we do for directives. The constraints will be laid down elsewhere in ACIR gen

TomAFrench:
Closing this PR as the `Oracle` opcode no longer exists.

PR #90, opened 2023-03-24
chore!: update to ACVM 0.7.0

TomAFrench:
This PR updates us to use ACVM 0.7.0.

Two things to take note of are
- `solve_black_box_func_call` must never return `OpcodeResolutionError::OpcodeNotSolvable`. I can't find any references to this in this repo so this should be safe.
- I'm not entirely sure how we want to handle serialisation of the new opcodes so I've added them to the TODO which we have for block opcodes.

### DIFF
--- a/common/src/serializer.rs
+++ b/common/src/serializer.rs
@@ -341,8 +341,8 @@ pub fn serialize_circuit(circuit: &Circuit) -> ConstraintSystem {
             Opcode::Directive(_) => {
                 // Directives are only needed by the pwg
             }
-            Opcode::Block(_, _) => {
-                // TODO: implement serialization of blocks to match BB's interface
+            Opcode::Block(_) | Opcode::RAM(_) | Opcode::ROM(_) | Opcode::Oracle(_) => {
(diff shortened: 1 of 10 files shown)


common/src/serializer.rs
```diff
@@ -341,16 +341,16 @@ pub fn serialize_circuit(circuit: &Circuit) -> ConstraintSystem {
             Opcode::Directive(_) => {
                 // Directives are only needed by the pwg
             }
-            Opcode::Block(_, _) => {
-                // TODO: implement serialization of blocks to match BB's interface
```
kevaundray, line 345: Barretenberg doesn't have these opcodes, so I think just saying they are unsupported is enough to note serialize it for now. Thry should get changed in the VM to opcodes that are supported by BB.

joss-aztec: I'd move `Opcode::Oracle` to the same case as `Opcode::Directive`. It's not a case of being unsupported - it simply doesn't lay constraints.

joss-aztec (approved):

PR #91, opened 2023-03-24
feat: store transcript in `.nargo/backends` directory

### LINKED ISSUE #53: Rename `noir_cache` or use a new location
The `noir_cache` directory is a bit weird, as it's not a dotfile. We should either rename to `.noir_cache` or maybe even put a `noir` entry in `~/.cache` (at least on mac?)

TomAFrench:
closes #53


common/src/crs.rs
```diff
@@ -5,14 +5,17 @@ const G1_START: usize = 28;
 const G2_START: usize = 28 + (5_040_001 * 64);
 const G2_END: usize = G2_START + 128 - 1;
 
+const BACKEND_IDENTIFIER: &str = "aztec_barretenberg_ultraplonk";
```
phated, line 8:
```suggestion
const BACKEND_IDENTIFIER: &str = "barretenberg_ultraplonk";
```
Are you good with removing aztec or do you think there will be "other barretenbergs"?

phated (approved):

common/src/crs.rs
```diff
@@ -5,15 +5,17 @@ const G1_START: usize = 28;
 const G2_START: usize = 28 + (5_040_001 * 64);
 const G2_END: usize = G2_START + 128 - 1;
 
+const BACKEND_IDENTIFIER: &str = "acvm-backend-barretenberg";
 const TRANSCRIPT_NAME: &str = "monomial-transcript00.dat";
```
phated, line 9:
```suggestion
const TRANSCRIPT_NAME: &str = "transcript00.dat";
```
Since we have a new directory for UP, let's revert my renaming here.

PR #94, opened 2023-03-30
feat: update `aztec_backend_wasm` to use new serialization

TomAFrench:
I've removed all of the circuit execution code as this will be the responsibility of the noir/nargo wasm crates. `aztec_backend_wasm` will just be responsible for converting ACIR into a format which BB can parse.


kobyhallx (approved):
💯

PR #95, opened 2023-04-03
chore: reenable commented out test cases

TomAFrench:
We're throwing warnings due to unused variables in the tests. I've then reenabled these tests where possible and dealt with the warnings where the tests fail. (I don't have context on why these test cases fail but I assume that they were commented out due to failing.)

I'll add an issue for the failing tests once this is merged.

### DIFF
--- a/barretenberg_static_lib/src/composer.rs
+++ b/barretenberg_static_lib/src/composer.rs
@@ -331,9 +331,7 @@ mod test {
             public_inputs: Assignments(vec![Scalar::one(), 3_i128.into()]),
             result: false,
         };
-        let test_cases = vec![
-            /*case_1,*/ case_2, case_3, /*case_4,*/ case_5, case_6,
-        ];
+        let test_cases = vec![case_1, case_2, case_3, case_4, case_5, case_6];
 
         test_composer_with_pk_vk(constraint_system, test_cases);
     }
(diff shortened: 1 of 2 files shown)


phated (changes requested):
Blocking this on the UP changes because they overlap (and we need green tests to verify it it working).

TomAFrench:
> Blocking this on the UP changes because they overlap (and we need green tests to verify it it working).

I don't quite follow on this. Looks like #84 disables the warnings but it doesn't re-enable the passing test cases in `barretenberg_static_lib` (which would help verify that it works).

barretenberg_static_lib/src/composer.rs
```diff
@@ -331,9 +331,7 @@ mod test {
             public_inputs: Assignments(vec![Scalar::one(), 3_i128.into()]),
             result: false,
         };
-        let test_cases = vec![
-            /*case_1,*/ case_2, case_3, /*case_4,*/ case_5, case_6,
-        ];
+        let test_cases = vec![case_1, case_2, case_3, case_4, case_5, case_6];
```
phated, line 334: I looked into these and something weird happens because they print `Serialized transcript does not contain the required number of bytes` which is a `throw_or_abort` error message, which makes bb abort in native. This is the wrong behavior and I believe we should keep these disabled until we properly solve in bb.

phated: This is also the reason they don't work in wasm, because `throw` is an unreachable trap and crashes the entire wasm instance.

PR #96, opened 2023-04-03
feat: add `decompress_witness_map` function

TomAFrench:
This PR adds the `decompress_witness_map` function which should be explained by its doc comment.


aztec_backend_wasm/src/lib.rs
```diff
@@ -1,14 +1,35 @@
...
 
-    let circuit = Circuit::read(&*acir_bytes).unwrap();
+    let circuit = Circuit::read(&*acir_bytes).expect("could not deserialize circuit");
     serialize_circuit(&circuit).to_bytes()
 }
+
+/// Returns the uncompressed flattened witness format expected by Barretenberg.
+///
+/// This function is necessary as witness maps are saved to disk in a compressed format;
+/// Barretenberg doesn't expect a compressed witness map however so it must be converted.
```
phated, line 20: @TomAFrench Should this "decompression" be the responsibility of *each* backend or noir/nargo? It seems like most backends won't want to deal with a compressed format.

phated (changes requested):
Big question

aztec_backend_wasm/src/lib.rs
TomAFrench: Yes, this was something I was thinking about as well. Decompression should be generic (as we're just undoing some processing done by `nargo` however we can't be sure that all backends will want the witness flattened in this way.

I think that while `ProofSystemCompiler::prove_with_pk` accepts the witness as a `BTreeMap<Witness, FieldElement>` then we should treat this as the responsibility of the backend. At least while when proving/verifying in the browser we're using a BB-specific npm package, we could probably handle this more nicely once we move to Joss's package.

TomAFrench: We have precedent for this being here as this function is basically equivalent to `packed_witness_to_witness` which was removed a little while back.

TomAFrench: The compile function returns the noir program's circuit rather than its witness so it's related to the current goal where we're wanting to prove/verify against an existing circuit.

While we wouldn't have to do this work if we were proving against a witness which has been generated within javascript, that adds a lot of extra work (abi encoding params, executing the circuit, etc.) so it's easier to test using pregenerated witnesses.

TomAFrench:
This function is equivalent to this function which was removed in #94 
https://github.com/noir-lang/aztec_backend/blob/28f1171a17452eec63a7e21e9c9e4e6f2ed6f3d2/aztec_backend_wasm/src/lib.rs#L65-L84

TomAFrench:
Closing as this package no longer exists.

PR #109, opened 2023-04-11
fix: enable `getrandom/js` with `js` feature flag

TomAFrench:
Not having this feature flag enabled prevents us from using `barretenberg_wasm` from within a webassembly context.

https://docs.rs/getrandom/latest/getrandom/#webassembly-support


TomAFrench:
Closing as Blaine is working on the same thing.

PR #112, opened 2023-04-12
fix: reorganize and ensure contracts can be compiled in Remix

phated:
This fixes some problems I encountered when copying the contract into Remix IDE. 

This is also waiting on responses on the [three items I opened on aztec-verifier-contracts](https://github.com/AztecProtocol/aztec-verifier-contracts/issues?q=is%3Aopen+author%3Aphated).


TomAFrench (approved):

TomAFrench:
This just needs a fresh pull of the verifier contracts and then should be gtg.

PR #117, opened 2023-04-17
feat!: merge `barretenberg_static_lib` and `barretenberg_wasm`

TomAFrench:
This PR starts the work of merging this workspace into a single crate. To simplify things, I've restricted the public interface of the crate to just be the `Plonk` struct (as this is necessary for nargo) to avoid any differences in the interface between targets. We should decide what else we want to expose before merging this.

The interfaces of most methods are identical between native and wasm backends. In these cases I've copied across the implementation from `barretenberg_wasm` and placed it in inside the function body from  `barretenberg_static_lib`, these are then switched between with feature flags.

 In a couple of cases the struct interfaces are different (`Barretenberg/StandardComposer/Pippenger`) so I've had to swap them out entirely based on the feature flags.


acvm_backend_barretenberg/Cargo.toml
```diff
@@ -22,8 +28,16 @@ blake2 = "0.9.1"
...
-default = [
-    # "acvm/bn254",
+default = ["native"]
+
+native=[
+    "dep:barretenberg-sys",
+    "common/std",
+]
+wasm = [
+    "wasmer",
```
phated, line 38:
```suggestion
    "dep:wasmer",
```

acvm_backend_barretenberg/Cargo.toml
```diff
@@ -32,7 +46,8 @@ default = [
     "wasmer/default-universal",
 ]
 js = [
-
-    # "acvm/bn254", 
+    "wasmer",
```
phated, line 49:
```suggestion
    "dep:wasmer",
```

phated (commented):
Do these need to specify `dep:`?

TomAFrench:
> Do these need to specify dep:?

I initially had `dep:` on here but that means we run into this error:

```shell
error: Package `acvm-backend-barretenberg v0.1.0 (/home/tom/Programming/aztec_backend/acvm_backend_barretenberg)` does not have feature `wasmer`.
It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

So it seems like having a `dep:` flag conflicts with switching on feature flags on that dependency. (adding `dep:` to the other wasmer flags also results in an error.)

https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

TomAFrench:
Tomorrow I'll probably break out the different flavours of `src/lib.rs` into separate files for native/wasm backends. It's pretty messy with the chonky if-else.

acvm_backend_barretenberg/build.rs
```diff
@@ -0,0 +1,31 @@
...
+
+// Useful for printing debugging messages during the build
+// macro_rules! p {
+//     ($($tokens: tt)*) => {
+//         println!("cargo:warning={}", format!($($tokens)*))
+//     }
+// }
+
+fn main() -> Result<(), String> {
+    let native_backend = env::var("CARGO_FEATURE_NATIVE").is_ok();
```
TomAFrench, line 11: Let's enforce mutually exclusive features here.

TomAFrench:
We're currently only testing the native backend due to us merging these two crates. I'd like to pass in the different feature flags so we can build/test the flake with both backends as otherwise we'll not be testing the wasm backend.

acvm_backend_barretenberg/Cargo.toml
```diff
@@ -22,8 +28,16 @@ blake2 = "0.9.1"
 tempfile = "*"
 
 [features]
-default = [
-    # "acvm/bn254",
+default = ["native"]
+
+native=[
+    "dep:barretenberg-sys",
```
kevaundray, line 34: why do you need "dep:"?

acvm_backend_barretenberg/src/pedersen.rs
```diff
@@ -0,0 +1,155 @@
...
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "native")] {
+                let result_bytes = barretenberg_sys::pedersen::compress_native(
+                    left.to_be_bytes().as_slice().try_into().unwrap(),
+                    right.to_be_bytes().as_slice().try_into().unwrap(),
+                );
+            } else {
+                use wasmer::Value;
+
+                let lhs_ptr = self.allocate(&left.to_be_bytes()); // 0..32
```
kevaundray, line 20: I don't think these comments are correct anymore since allocate will not place 32 bytes, starting from pointer address 0. `lhs_ptr` and `rhs_ptr` are allocated somewhere.

With that in mind, you could do `let result_ptr = Value::I32(0);` and then `let result_bytes = self.slice_memory(0, 32);`

acvm_backend_barretenberg/src/scalar_mul.rs
```diff
@@ -1,15 +1,25 @@
...
+    pub(crate) fn fixed_base(&mut self, input: &FieldElement) -> (FieldElement, FieldElement) {
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "native")] {
+                use super::field_to_array;
+
+                let result_bytes = barretenberg_sys::schnorr::construct_public_key(&field_to_array(input));
+            } else {
+                use wasmer::Value;
+
+                let lhs_ptr = self.allocate(&input.to_be_bytes()); // 0..32
```
kevaundray, line 15: Same comment regarding 0..32 not being the memory location of `lhs_ptr`

kevaundray (commented):
Left some general comments; was reading to understand more about this issue: https://github.com/noir-lang/acvm/issues/211

acvm_backend_barretenberg/Cargo.toml
TomAFrench: Some discussion of this above: https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

TomAFrench:
[8156f62](https://github.com/noir-lang/aztec_backend/pull/117/commits/8156f621b19f027245e68aa78674ef73dc0ddd85) takes some of the insight from https://github.com/noir-lang/aztec_backend/pull/127 and applies it here.

I've verified that we can still have `Barretenberg` be immutable while caching the CRS on it by using a `RefCell` however I'm not sure on the correct way to reconcile that with the usage of `std::mem::forget(g2_data)` so I've erred on just loading a fresh copy each time and maintaining the `forget`. We can revisit this later on.

acvm_backend_barretenberg/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+
+impl Default for Barretenberg {
+    fn default() -> Barretenberg {
+        Barretenberg::new()
+    }
+}
+
+impl Barretenberg {
+    fn get_crs(&mut self, num_points: u32) -> CRS {
+        // TODO: yucky, we can avoid this cloning.
```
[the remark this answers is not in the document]
TomAFrench: Yeah this is due to cloning the CRS string. Ideally we wouldn't have to copy those large vectors if we can avoid it.

acvm_backend_barretenberg/Cargo.toml
```diff
@@ -22,8 +29,16 @@ blake2 = "0.9.1"
 tempfile = "*"
 
 [features]
-default = [
-    # "acvm/bn254",
+default = ["wasm"]
```
phated, line 32: I'm playing around with this locally and I'm seeing build failures when I turn this to `"native"` so I think this needs some more work.

acvm_backend_barretenberg/src/scalar_mul.rs
```diff
@@ -1,21 +1,30 @@
...
-use wasmer::Value;
 
 use super::{Barretenberg, FIELD_BYTES};
 
 impl Barretenberg {
-    pub fn fixed_base(&mut self, input: &FieldElement) -> (FieldElement, FieldElement) {
-        let lhs_ptr: usize = 0;
-        let result_ptr: usize = lhs_ptr + FIELD_BYTES;
+    pub(crate) fn fixed_base(&self, input: &FieldElement) -> (FieldElement, FieldElement) {
+        cfg_if::cfg_if! {
```
phated, line 7: I'm getting super lost in all the cfg_if branches. Can we just make the various implementations their own functions and just toggle them with the config macro?

phated (changes requested):
Some things I noticed when I pulled this down locally.

acvm_backend_barretenberg/Cargo.toml
TomAFrench: Oops, looks like this snuck in when merging #133 back in. Fixed.

TomAFrench:
Ah, this still needs the multiple outputs (native vs wasm) added to the flake as well.

acvm_backend_barretenberg/src/scalar_mul.rs
TomAFrench: My concern about this was ensuring that we match interfaces. I've used private traits to address this in [1d2369e](https://github.com/noir-lang/aztec_backend/pull/117/commits/1d2369e0e91db1581452f3e242347e294d0c23d0)

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -74,24 +85,32 @@ impl StandardComposer {
...
-    pub fn compute_verification_key(&self, proving_key: &[u8]) -> Vec<u8> {
+    fn compute_verification_key(
+        &self,
+        constraint_system: &ConstraintSystem,
+        proving_key: &[u8],
+    ) -> Vec<u8> {
+        let circuit_size = self.get_circuit_size(constraint_system);
+        let CRS {
+            g1_data, g2_data, ..
+        } = CRS::new(circuit_size as usize);
```
phated, line 96: Does this actually use g1_data? Because it might be more efficient to use a `G2::new` instead

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -101,30 +120,37 @@ impl StandardComposer {
...
+        &self,
+        constraint_system: &ConstraintSystem,
         witness: WitnessAssignments,
         proving_key: &[u8],
     ) -> Vec<u8> {
-        let cs_buf = self.constraint_system.to_bytes();
+        let circuit_size = self.get_circuit_size(constraint_system);
+        let CRS {
+            g1_data, g2_data, ..
+        } = CRS::new(circuit_size as usize);
```
phated, line 132: Does this actually use g1_data? Because it might be more efficient to use a `G2::new` instead

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -133,31 +159,36 @@ impl StandardComposer {
...
+        &self,
+        constraint_system: &ConstraintSystem,
         // XXX: Important: This assumes that the proof does not have the public inputs pre-pended to it
         // This is not the case, if you take the proof directly from Barretenberg
         proof: &[u8],
         public_inputs: Assignments,
         verification_key: &[u8],
     ) -> bool {
+        let circuit_size = self.get_circuit_size(constraint_system);
+        let CRS { g2_data, .. } = CRS::new(circuit_size as usize);
```
phated, line 176: Does this actually use g1_data? Because it might be more efficient to use a `G2::new` instead

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -167,8 +198,235 @@ impl StandardComposer {
...
+        constraint_system: &ConstraintSystem,
+        proving_key: &[u8],
+    ) -> Vec<u8> {
+        use super::POINTER_BYTES;
+        use wasmer::Value;
+
+        let circuit_size = self.get_circuit_size(constraint_system);
+        let CRS {
+            g1_data, g2_data, ..
+        } = CRS::new(circuit_size as usize);
```
phated, line 282: Does this actually use g1_data? Because it might be more efficient to use a G2::new instead

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -167,8 +198,235 @@ impl StandardComposer {
...
+        witness: WitnessAssignments,
+        proving_key: &[u8],
+    ) -> Vec<u8> {
+        use super::POINTER_BYTES;
+        use wasmer::Value;
+
+        let circuit_size = self.get_circuit_size(constraint_system);
+        let CRS {
+            g1_data, g2_data, ..
+        } = CRS::new(circuit_size as usize);
```
phated, line 326: Does this actually use g1_data? Because it might be more efficient to use a G2::new instead

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -167,8 +198,235 @@ impl StandardComposer {
...
+        // XXX: Important: This assumes that the proof does not have the public inputs pre-pended to it
+        // This is not the case, if you take the proof directly from Barretenberg
+        proof: &[u8],
+        public_inputs: Assignments,
+        verification_key: &[u8],
+    ) -> bool {
+        use wasmer::Value;
+
+        let circuit_size = self.get_circuit_size(constraint_system);
+        let CRS { g2_data, .. } = CRS::new(circuit_size as usize);
```
phated, line 378: Using G2::new would be more efficient

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -167,8 +198,235 @@ impl StandardComposer {
...
+        }
+    }
+}
+
 fn pow2ceil(v: u32) -> u32 {
-    v.next_power_of_two()
+    #[cfg(feature = "native")]
+    {
+        v.next_power_of_two()
+    }
```
phated, line 420: Why use this helper at all? Seems like the `not(native)` impl would work for both

acvm_backend_barretenberg/src/lib.rs
```diff
@@ -0,0 +1,331 @@
+#![warn(unused_crate_dependencies, unused_extern_crates)]
+#![warn(unreachable_pub)]
+
+// `acvm-backend-barretenberg` can either interact with the Barretenberg backend through a static library
+// or through an embedded wasm binary. It does not make sense to include both of these backends at the same time.
+// We then throw a compilation error if both flags are set.
+// TODO: handle JS target.
+#[cfg(all(feature = "native", feature = "wasm"))]
+compile_error!("feature \"native\"  and feature \"wasm\" cannot be enabled at the same time");
```
phated, line 9:
```suggestion
compile_error!("feature \"native\" and feature \"wasm\" cannot be enabled at the same time");
```
nit: remove space

acvm_backend_barretenberg/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+
+mod acvm_interop;
+mod blake2s;
+mod composer;
+#[cfg(all(feature = "native", test))]
```
phated, line 14: This is in native and wasm, right? It just doesn't work in JS right now

acvm_backend_barretenberg/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+cfg_if::cfg_if! {
+    if #[cfg(feature = "native")] {
+
+        impl Barretenberg {
+            pub(crate) fn new() -> Barretenberg {
+                Barretenberg {}
+            }
+        }
+
+        fn field_to_array(f: &common::acvm::FieldElement) -> [u8; 32] {
```
phated, line 58: I'm surprised this wouldn't be used in wasm, are we hand-rolling it or something?

acvm_backend_barretenberg/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+
+                #[cfg(not(feature = "js"))]
+                {
+                    for (byte_id, cell) in memory.uint8view()[offset..(offset + arr.len())]
+                        .iter()
+                        .enumerate()
+                    {
+                        cell.set(arr[byte_id]);
+                    }
+                }
```
phated, line 137: For a diff PR: I want to explore if this can be written with an impl that supports JS and wasmer

acvm_backend_barretenberg/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+                let end = start + length;
+
+                #[cfg(feature = "js")]
+                return memory.uint8view().to_vec()[start as usize..end].to_vec();
+
+                #[cfg(not(feature = "js"))]
+                return memory.view()[start..end]
+                    .iter()
+                    .map(|cell: &Cell<u8>| cell.get())
+                    .collect();
```
phated, line 151: Same as above

acvm_backend_barretenberg/src/pedersen.rs
```diff
@@ -1,12 +1,59 @@
...
+    fn compress_native(&self, left: &FieldElement, right: &FieldElement) -> FieldElement {
+        let result_bytes = barretenberg_sys::pedersen::compress_native(
+            left.to_be_bytes().as_slice().try_into().unwrap(),
+            right.to_be_bytes().as_slice().try_into().unwrap(),
+        );
+
+        FieldElement::from_be_bytes_reduce(&result_bytes)
+    }
+
+    #[allow(dead_code)]
```
phated, line 22: ?

acvm_backend_barretenberg/src/pedersen.rs
```diff
@@ -27,7 +74,12 @@ impl Barretenberg {
         FieldElement::from_be_bytes_reduce(&result_bytes)
     }
 
-    pub fn compress_many(&mut self, inputs: Vec<FieldElement>) -> FieldElement {
+    #[allow(dead_code)]
```
phated, line 77: ?

acvm_backend_barretenberg/src/schnorr.rs
```diff
@@ -33,12 +65,17 @@ impl Barretenberg {
         let sig_s: [u8; 32] = sig_s_bytes.try_into().unwrap();
         let sig_e: [u8; 32] = sig_e_bytes.try_into().unwrap();
 
-        [sig_s, sig_e].concat().try_into().unwrap()
+        let sig_bytes: [u8; 64] = [sig_s, sig_e].concat().try_into().unwrap();
+        sig_bytes
     }
 
-    pub fn construct_public_key(&mut self, private_key: [u8; 32]) -> [u8; 64] {
+    #[allow(dead_code)]
```
phated, line 72: ?

flake.nix
```diff
@@ -153,14 +153,22 @@
...
+        cargo-test-native = craneLib.cargoTest (commonArgs // {
           inherit cargoArtifacts;
 
           cargoTestExtraArgs = "--workspace";
 
           # It's unclear why doCheck needs to be enabled for tests to run but not clippy
           doCheck = true;
         });
+
+        cargo-test-wasm = craneLib.cargoTest (commonArgs // {
```
phated, line 165: Do we need a clippy with the wasm feature enabled?

phated (changes requested):
I noticed a few more things. I wonder if we can drop the cfg-if block for all the setup but I won't block on it.

acvm_backend_barretenberg/Cargo.toml
```diff
@@ -9,10 +9,17 @@ rust-version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-wasmer = { version = "*", default-features = false }
+blake2 = "0.9.1"
+cfg-if = "1.0.0"
```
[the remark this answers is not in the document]
phated: Not gone yet.

acvm_backend_barretenberg/Cargo.toml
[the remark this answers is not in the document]
phated: Since this isn't native, both nix test runs are testing wasm

acvm_backend_barretenberg/src/acvm_interop/proof_system.rs
```diff
@@ -62,12 +60,16 @@ impl ProofSystemCompiler for Plonk {
...
         verification_key: &[u8],
     ) -> bool {
-        let mut composer = StandardComposer::new(circuit.into());
-
         // Unlike when proving, we omit any unassigned witnesses.
         // Witness values should be ordered by their index but we skip over any indices without an assignment.
         let flattened_public_inputs: Vec<FieldElement> = public_inputs.into_values().collect();
 
-        composer.verify_with_vk(proof, flattened_public_inputs.into(), verification_key)
+        Composer::verify_with_vk(
```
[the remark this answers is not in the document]
TomAFrench: This is due to name collisions with the `Backend` trait, Rust can't figure out which method we want to call so we need to be explicit.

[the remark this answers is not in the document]
phated: I'm happy to change the internal name, I've not been a fan of them having conflicting names. It probably should be `verify_constraint_system_with_vk` since it takes the CS instead of Circuit.

acvm_backend_barretenberg/src/lib.rs
TomAFrench: The first of these tests require native features to run and both will quickly be invalidated as we change out how we handle the CRS. I'm inclined to leave it as is and we then fix it properly as we're writing that new code.

acvm_backend_barretenberg/src/lib.rs
TomAFrench: This is due to the fact that we need to pass `[u8;32]` to barretenberg_sys but we can just pass `&[u8]` when writing to wasm memory.

acvm_backend_barretenberg/src/pedersen.rs
TomAFrench: The code was already dead, we're just not making it public anymore. (causing the lint)

acvm_backend_barretenberg/src/lib.rs
phated: Ah, this is only for the tests, ignore me then

acvm_backend_barretenberg/src/pedersen.rs
phated: Where are you seeing this? I removed all the dead code annotations and I'm not getting any clippy alerts.

TomAFrench:
>  I wonder if we can drop the cfg-if block for all the setup but I won't block on it.

I don't see a good way to do this. Rust doesn't seem to like putting a scope around the two branches (similarly to how we do in `pow2ceil`) so we'd have to add the `cfg` to each and every element inside which would be uglier.

phated:
> I don't see a good way to do this. Rust doesn't seem to like putting a scope around the two branches (similarly to how we do in pow2ceil) so we'd have to add the cfg to each and every element inside which would be uglier.

I'm wondering if it would be better/worse to use your cfg on a mod like you did for crs.rs?

acvm_backend_barretenberg/src/pedersen.rs
TomAFrench: image

TomAFrench: It'll only appear on the version corresponding to the feature flag currently being compiled. Maybe you removed it from an inactive block?

TomAFrench:
You asked me to revert that change though :melting_face:

Yeah, I'm happy with that as it encapsulates more of the wasm internals.

acvm_backend_barretenberg/src/pedersen.rs
phated: Yeah, I'm not sure. I suppose we can leave them

acvm_backend_barretenberg/src/composer.rs
TomAFrench: Sure, this is all private so swapping out the implementation later is nbd.

Let's avoid doing so many checks though as we know the value for `v` which will begin to overflow

```rust
if v > (u32::MAX >> 1) {
    panic!("pow2ceil overflow");
}

let mut p = 1;
while p < v {
    p <<= 1;
}
p
```

phated: sounds good to me 🎉

phated (approved):
This is awesome! Let's get it merged so we can keep iterating on this crate 🎉

PR #120, opened 2023-04-17
feat!: Implement pseudo-builder pattern for ConstraintSystem & hide struct fields

phated:
This builds upon #118 and hides all fields on the ConstraintSystem struct. It then implements an Builder-like pattern on the ConstraintSystem for building them up in tests (note that this is only ever used in tests and in normal use these are constructed via `From<&Circuit>`)

I opted away from an entirely new struct for the builder pattern because we use vecs for everything and they can be defaulted to empty vecs. Marking as a draft so we can decide if this is good enough.

### DIFF
--- a/barretenberg_static_lib/src/acvm_interop/smart_contract.rs
+++ b/barretenberg_static_lib/src/acvm_interop/smart_contract.rs
@@ -43,21 +43,10 @@ fn test_smart_contract() {
         qc: Scalar::zero(),
     };
 
-    let constraint_system = ConstraintSystem {
-        var_num: 4,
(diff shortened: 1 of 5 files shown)


TomAFrench:
> note that this is only ever used in tests and in normal use these are constructed via From<&Circuit>

Shall we add `#[cfg(test)]` to these functions in this case to avoid them hitting the external interface?

phated:
I can add a comment to make them `#[cfg(test)]` once merged into one crate

TomAFrench (approved):
LGTM

PR #121, opened 2023-04-17
fix: use `Barretenberg.call` to query circuit size from wasm

TomAFrench:
It seems like someone accidentally reimplemented `Barretenberg.call` when writing these functions. I've then replaced it with just using the proper method.

### DIFF
--- a/barretenberg_wasm/src/composer.rs
+++ b/barretenberg_wasm/src/composer.rs
@@ -52,24 +52,15 @@ impl StandardComposer {
         let cs_buf = constraint_system.to_bytes();
         let cs_ptr = barretenberg.allocate(&cs_buf);
 
-        let func = barretenberg
-            .instance
-            .exports
-            .get_function("acir_proofs_get_total_circuit_size")
-            .unwrap();
+        let result_bytes = barretenberg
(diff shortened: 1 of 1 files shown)


phated (dismissed):

barretenberg_wasm/src/composer.rs
```diff
@@ -79,24 +70,15 @@ impl StandardComposer {
...
-        }
+        let circuit_size = barretenberg
+            .call("acir_proofs_get_exact_circuit_size", &cs_ptr)
+            .into_i32();
+        let circuit_size =
+            u32::try_from(circuit_size).expect("circuit cannot have negative number of gates");
+
+        barretenberg.free(cs_ptr);
+
+        pow2ceil(circuit_size + StandardComposer::NUM_RESERVED_GATES)
```
phated, line 81: I think this is different than the above and doesn't use the pow2ceil and doesn't add the reserved

phated (changes requested):
One change is needed, I think

barretenberg_wasm/src/composer.rs
TomAFrench: Ah, my bad. Fixing now

kobyhallx (approved):
lgtm 👍

PR #123, opened 2023-04-18
feat!: return boolean rather than `FieldElement` from `verify_signature`

TomAFrench:
This PR simplifies the `verify_signature` function so that it returns a boolean for whether the signature was valid or not rather than the field element which will be assigned to a witness as a result of this.

This translation is now done in `solve_black_box_func_call`.


jfecher (approved):
👍

PR #124, opened 2023-04-18
feat: return boolean rather than `FieldElement` from `check_membership`

TomAFrench:
Similar to #123 but for a different method.


phated (approved):
These get cleaned up even further with 0.9.1 of ACVM, right?

PR #127, opened 2023-04-20
chore: Remove mut throughout interfaces

phated:
While exploring https://github.com/noir-lang/acvm/issues/211, I wanted to see exactly **why** this backend needed things as `&mut self` so I started removing stuff. It seems like the only things that needed to be mutable were Assignments and the Blake2s hasher. These could both be worked-around by using `std::cell::RefCell`, I believe.

This seems to pass all the tests and remove mut on pretty much the entire interface (unless my grepping missed something).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,7 +267,6 @@ name = "barretenberg_static_lib"
 version = "0.1.0"
 dependencies = [
  "barretenberg-sys",
- "blake2",
  "common",
  "tempfile",
 ]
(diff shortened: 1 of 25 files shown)


TomAFrench:
Ahhh, this is really nice. I'm not sure why but I was thinking something along the lines of "when we're interacting with allocating/freeing memory in wasm we're mutating the `Memory` field of `Barretenberg` so we need a mutable reference." so didn't challenge the need for it.

In hindsight, dealing with the memory inside wasm is kinda outside the purview of the borrow checker (hinted by the fact that `Memory` itself is listed as immutable.)

TomAFrench:
I don't think that we need to make the changes to `Assignments` and the Blake2s hasher as these are purely internal details so we can just have them be mutable.

phated:
Sounds good. I believe since you took the majority of these changes in #117 that we're just going to close this?

PR #130, opened 2023-04-21
Update barretenberg-sys to 0.1.3

kevaundray:
This will allow us to remove the global test_threads=1 setting in .github which should also speed up other crates like acvm.


phated (changes requested):
This needs nix changes too, but I'm updating this stuff now and multi-threaded causes composer tests to fail.

kevaundray:
> This needs nix changes too, but I'm updating this stuff now and multi-threaded causes composer tests to fail.

Hmm will need to check for further thread-unsafe code

TomAFrench:
Do we need to wrap each test individually? The docs seem like you can wrap the entire suite in one go.

PR #131, opened 2023-04-21
chore: Bring project up to date with Nix patterns

phated:
This should be everything we need to bring aztec_backend up-to-date with the Nix & direnv patterns in bb-sys and Noir.


flake.nix
```diff
@@ -112,40 +150,39 @@
         cargo-clippy = craneLib.cargoClippy (commonArgs // {
           inherit cargoArtifacts;
 
-          cargoClippyExtraArgs = "--all-targets --workspace -- -D warnings";
-
-          doCheck = true;
+          cargoClippyExtraArgs = "--all-targets -- -D warnings";
```
TomAFrench, line 153:
```suggestion
          cargoClippyExtraArgs = "--all-targets --workspace -- -D warnings";
```

We should still run clippy over the whole workspace for now.

TomAFrench (commented):
Looks good. Can you update cspell and then we can merge?

flake.nix
phated: I tested and I'm pretty sure this is how `--all-targets` works.

TomAFrench: image
This doesn't seem to relate to workspaces though

phated: > By default, when no package selection options are given, the packages selected depend on the selected manifest file (based on the current working directory if --manifest-path is not given). If the
       manifest is the root of a workspace then the workspaces default members are selected, otherwise only the package defined by the manifest will be selected.

Seems that it selects all our members and that this was only something to think about when we had excluded members. Yeah?

TomAFrench: Yeah, I guess it wouldn't pick a random package if we don't have a default so it must be doing all of them atm.

TomAFrench (approved):

phated:
I updated the branch protections

PR #133, opened 2023-04-24
chore: replace hardcoded pointer locations with named values

TomAFrench:
Addresses the following review comments:

https://github.com/noir-lang/aztec_backend/pull/117#discussion_r1172624066
https://github.com/noir-lang/aztec_backend/pull/117#discussion_r1172639691
https://github.com/noir-lang/aztec_backend/pull/117#discussion_r1172641210

I've broken these out of the main PR as they're a helpful readability change by themselves and we can merge master back into that branch.

Schnorr tests are failing so I need to nail that down.

### DIFF
--- a/barretenberg_wasm/src/blake2s.rs
+++ b/barretenberg_wasm/src/blake2s.rs
@@ -1,4 +1,4 @@
-use super::Barretenberg;
+use super::{Barretenberg, FIELD_BYTES};
 use common::acvm::FieldElement;
 use wasmer::Value;
 
@@ -6,20 +6,21 @@ impl Barretenberg {
     // TODO : Replace this call with a blake2s call and a field element reduce
     /// Hashes to a bn254 scalar field element using blake2s
     pub fn hash_to_field(&mut self, input: &[u8]) -> FieldElement {
-        let input_ptr = self.allocate(input); // 0..32
+        let input_ptr = self.allocate(input);
+        let result_ptr: usize = 0;
 
-        let result_ptr = Value::I32(0);
-
-        // Not sure why this is needed to send to WASM
-        // It seems to be sent twice?
-        let data_len = Value::I32(input.len() as i32);
-
-        self.call_multiple("blake2s_to_field", vec![&input_ptr, &data_len, &result_ptr]);
+        self.call_multiple(
+            "blake2s_to_field",
+            vec![
+                &input_ptr,
+                &Value::I32(input.len() as i32),
+                &Value::I32(result_ptr as i32),
+            ],
+        );
 
         self.free(input_ptr);
 
-        // result_ptr is 0 and the output is 32 bytes long
-        let result_bytes = self.slice_memory(0, 32);
+        let result_bytes = self.slice_memory(result_ptr, result_ptr + FIELD_BYTES);
         FieldElement::from_be_bytes_reduce(&result_bytes)
     }
 }
--- a/barretenberg_wasm/src/blake2s.rs
+++ b/barretenberg_wasm/src/blake2s.rs
@@ -20,7 +20,7 @@ impl Barretenberg {
 
         self.free(input_ptr);
 
-        let result_bytes = self.slice_memory(result_ptr, result_ptr + FIELD_BYTES);
+        let result_bytes = self.slice_memory(result_ptr, FIELD_BYTES);
         FieldElement::from_be_bytes_reduce(&result_bytes)
     }
 }
--- a/barretenberg_wasm/src/composer.rs
+++ b/barretenberg_wasm/src/composer.rs
@@ -85,21 +85,21 @@ impl StandardComposer {
         let cs_ptr = self.barretenberg.allocate(&cs_buf);
 
         // The proving key is not actually written to this pointer.
-        // `result_ptr` is a pointer to a pointer which holds the proving key.
-        let result_ptr: usize = 0;
+        // `pk_ptr_ptr` is a pointer to a pointer which holds the proving key.
+        let pk_ptr_ptr: usize = 0;
 
         let pk_size = self
             .barretenberg
             .call_multiple(
                 "acir_proofs_init_proving_key",
-                vec![&cs_ptr, &Value::I32(result_ptr as i32)],
+                vec![&cs_ptr, &Value::I32(pk_ptr_ptr as i32)],
             )
             .value();
         let pk_size: usize = pk_size.unwrap_i32() as usize;
 
-        // We then need to read the pointer at `result_ptr` to get the key's location
+        // We then need to read the pointer at `pk_ptr_ptr` to get the key's location
         // and then slice memory again at `pk_ptr` to get the proving key.
-        let pk_ptr = self.barretenberg.slice_memory(result_ptr, POINTER_BYTES);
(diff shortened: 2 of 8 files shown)


barretenberg_wasm/src/blake2s.rs
```diff
@@ -1,25 +1,26 @@
...
+                &Value::I32(input.len() as i32),
+                &Value::I32(result_ptr as i32),
+            ],
+        );
 
         self.free(input_ptr);
 
-        // result_ptr is 0 and the output is 32 bytes long
-        let result_bytes = self.slice_memory(0, 32);
+        let result_bytes = self.slice_memory(result_ptr, result_ptr + FIELD_BYTES);
```
kevaundray, line 23: Just a thought, I wonder if we should change the API slightly such that we don't need to do the arithmetic here, it would then be:

```rust
self.slice_memory(result_ptr, FIELD_BYTES);
```

i.e. instead of giving the absolute address for start and end, you just give the address and then an offset.

barretenberg_wasm/src/composer.rs
```diff
@@ -149,18 +168,23 @@ impl StandardComposer {
...
                 ],
             )
             .value();
+        let proof_size: usize = proof_size.unwrap_i32() as usize;
 
-        let proof_ptr = self.barretenberg.slice_memory(0, 4);
-        let proof_ptr = u32::from_le_bytes(proof_ptr[0..4].try_into().unwrap());
+        // We then need to read the pointer at `result_ptr` to get the proof's location
+        // and then slice memory again at `proof_ptr` to get the proof data.
+        let proof_ptr = self
```
kevaundray, line 179: would it be confusing to name this `proof_ptr_ptr` to denote the fact that its a pointer to a pointer, then L182 would  simply be `proof_ptr`

barretenberg_wasm/src/lib.rs
```diff
@@ -21,6 +21,10 @@ pub use common::crs;
 use std::cell::Cell;
 use wasmer::{imports, Function, Instance, Memory, MemoryType, Module, Store, Value};
 
+const FIELD_BYTES: usize = 32;
+const SIG_BYTES: usize = 64;
```
kevaundray, line 25: small nit: do we want to call this SCHNOR_SIG_BYTES since not all signatures are the same, or is this implied from the context?

kevaundray: comments on what these are would be good

barretenberg_wasm/src/pedersen.rs
```diff
@@ -3,45 +3,50 @@ use wasmer::Value;
...
 
-        let result_bytes = self.slice_memory(0, 32);
+        let result_bytes = self.slice_memory(result_ptr, result_ptr + FIELD_BYTES);
         FieldElement::from_be_bytes_reduce(&result_bytes)
     }
 
     pub fn encrypt(&mut self, inputs: Vec<FieldElement>) -> (FieldElement, FieldElement) {
         let input_buf = Assignments::from(inputs).to_bytes();
         let input_ptr = self.allocate(&input_buf);
+        let result_ptr: usize = 32;
```
kevaundray, line 39: not blocking, I think this can just be 0

barretenberg_wasm/src/pedersen.rs
```diff
@@ -3,45 +3,50 @@ use wasmer::Value;
...
-use super::Barretenberg;
+use super::{Barretenberg, FIELD_BYTES};
 
 impl Barretenberg {
     pub fn compress_native(&mut self, left: &FieldElement, right: &FieldElement) -> FieldElement {
-        let lhs_ptr = self.allocate(&left.to_be_bytes()); // 0..32
-        let rhs_ptr = self.allocate(&right.to_be_bytes()); // 32..64
-        let result_ptr = Value::I32(64); // 64..96
+        let lhs_ptr = self.allocate(&left.to_be_bytes());
+        let rhs_ptr = self.allocate(&right.to_be_bytes());
```
TomAFrench, line 11: Seems like we should be using `move_to_heap` here.

barretenberg_wasm/src/schnorr.rs
```diff
@@ -1,49 +1,73 @@
...
     pub fn construct_signature(&mut self, message: &[u8], private_key: [u8; 32]) -> [u8; 64] {
-        self.transfer_to_heap(&private_key, 64);
-        self.transfer_to_heap(message, 96);
-        let message_len = Value::I32(message.len() as i32);
+        let message_ptr: usize = 96;
+        let private_key_ptr: usize = 64;
+        let sig_s_ptr: usize = 0;
+        let sig_e_ptr: usize = 32;
+        let result_ptr: usize = 0;
+
```
kevaundray, line 18: can we add comments here to say why these numbers are the way they are? 

Also I think this function is slightly confusing if you stare at it long enough: 

The result of construct_signature are two field elements (s,e) , so it makes sense that we have two pointers `sig_s` and `sig_e` -- so then why do we have `result_ptr`. 

It would be a bit clearer if we did: `let result_ptr = sig_s_ptr`. This implicitly relies on the fact that we are storing `sig_s` and `sig_e` right next to each other, s is [0,32) and e is [32,64). 

I would probably just do the following:

```rust
  let sig_s_bytes = self.slice_memory(sig_s_ptr, sig_s_ptr + FIELD_BYTES);
  let sig_e_bytes = self.slice_memory(sig_e_ptr, sig_e_ptr + FIELD_BYTES);
```

Then just return (s, e) and in verify_signature we could pass it (s,e) and then there is no need to split

kevaundray: The change in API was only a suggestion though I think we should change `let result_ptr: usize = 0;` because it must always be equal to sig_s_ptr

barretenberg_wasm/src/schnorr.rs
```diff
@@ -1,49 +1,73 @@
...
+            .try_into()
+            .unwrap()
     }
 
     pub fn verify_signature(&mut self, pub_key: [u8; 64], sig: [u8; 64], message: &[u8]) -> bool {
-        self.transfer_to_heap(&pub_key, 0);
-        self.transfer_to_heap(&sig[0..32], 64);
-        self.transfer_to_heap(&sig[32..64], 96);
-        self.transfer_to_heap(message, 128);
+        let message_ptr: usize = 128;
```
kevaundray, line 56: similar to above I would change this to:

```rust
       let public_key_ptr: usize = 0;
       let sig_s_ptr: usize = public_key_ptr + pub_key.len();
       let sig_e_ptr: usize = sig_s_ptr + FIELD_BYTES;
       let message_ptr: usize = sig_e_ptr + FIELD_BYTES;
```

kevaundray (commented):
This looks good overall, I've added some comments

barretenberg_wasm/src/schnorr.rs
TomAFrench: I think for the interests of merging this back into `acvm-backend-barretenberg` I'm going to leave the interface the same as it is currently and we can address this up there.

kevaundray (approved):

phated (approved):
lgtm

PR #134, opened 2023-04-24
chore(ci): Cache cargo dependencies between runs

phated:
This caches the "cargoArtifacts" derivation from Nix via GitHub Actions Cache. That includes all cargo dependencies as produced by `craneLib.buildDepsOnly`. We then import the deps into Nix if the cache exists.

This reduced the CI run on ubuntu down to 7 minutes, roughly 5-6 of it being the actual test run.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: /tmp/nix-cache
-          key: ${{ runner.os }}-flake-${{ hashFiles('flake.lock') }}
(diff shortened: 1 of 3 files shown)


TomAFrench (commented):
Can we copy this into a github action which we can reuse across the organisation? I can see an identical PR in `noir-lang-noir`.

phated:
Yeah, need to make these reusable, but there's enough hacking that I need to do to test these changes that I want to be confident before we do that. Especially because every change in a reusable is reflected in every project.

TomAFrench:
Cool, yeah no point in getting ahead of ourselves.

PR #135, opened 2023-04-24
chore: Use pollster as async executor

phated:
This drops the direct `tokio` dependency (it still exists inside reqwest) and uses [pollster](https://crates.io/crates/pollster) for `block_on`. Their package readme has a good "why" for itself.

I was trying to understand if we can `block_on` when compiling to JS with wasm-bindgen and found this comment: https://github.com/async-rs/async-std/issues/913#issuecomment-1296282060 which recommended pollster. I tried it in a demo wasm-pack project I have an it seems to block on an async function, so I think we can layer JS promises and it should work 🤞

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,6 +269,7 @@ dependencies = [
  "barretenberg-sys",
  "blake2",
  "common",
+ "pollster",
  "tempfile",
 ]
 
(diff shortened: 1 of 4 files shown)


phated:
Converting to draft because it seems that `Condvar::wait` throws when interop'ing with JS promises.

phated:
Learned a lot here but closing this. We might use pollster to block_on async functions in something like nargo_cli.

Issue #136, opened 2023-04-25
Consider removing dead code from repo


TomAFrench:
As part of the tightening up of the interface done in https://github.com/noir-lang/aztec_backend/pull/117. I noticed that we don't actually use a decent number of the functions which we've written.

- ~~`hash_to_field` in `blake2s.rs`~~.
- `compress_many` in `pedersen.rs`.

We also don't use `construct_public_key` in `schnorr.rs` however that's useful for testing whereas the other two are not.

We should consider removing these functions if this package is being restricted to serve only the ACVM interface.

TomAFrench:
We've removed so much code from here that github thinks this is a solidity project.

PR #137, opened 2023-04-25
feat!: remove dead blake2 code

TomAFrench:
We don't make use of barretenberg to perform this hashing so we can remove it and continue using the rust implementation from ACVM instead.

https://github.com/noir-lang/aztec_backend/blob/7281750ae4d265f2d81a0eaad8a654d38133e1d5/common/src/black_box_functions.rs#L28

Alternatively, we could retain this just for testing that this implementation matches Barretenberg but that seems overkill to me.

Related to #136


kevaundray (approved):
Thanks for the CC @phated -- @TomAFrench yeah this looks correct, in common/black_box_functions, we use the rust version

phated (changes requested):
Actually, I want to hold this for @vezenovm's feedback because he might have used this function and test to get values from BB for tests.

vezenovm:
> Actually, I want to hold this for @vezenovm's feedback because he might have used this function and test to get values from BB for the composer tests.

It is ok this can be removed

phated (approved):

PR #138, opened 2023-04-25
chore: remove outdated comment

kevaundray:
This comment was from a time when each backend would define how it would do particular operations. Now, we are of the opinion that this should be standardized by ACIR


phated (approved):

TomAFrench:
@kevaundray Seems like we should be moving this implementation down into ACVM in that case similarly to `BlackBoxFunc::SHA256` and `BlackBoxFunc::Blake2s` in that case.

kevaundray:
> @kevaundray Seems like we should be moving this implementation down into ACVM in that case similarly to `BlackBoxFunc::SHA256` and `BlackBoxFunc::Blake2s` in that case?

Yep that sounds right

PR #139, opened 2023-04-26
feat!: Implement CommonReferenceString trait

### LINKED ISSUE #160: Remove filesystem operations from the backend
### Problem

Currently, this backend manages caching of its CRS, but it writes to a filepath that should be maintained by Nargo.

### Proposed solution

We should change the backend trait to return a CRS as bytes and Nargo can manage caching it. This will ensure that all CRS from various backends are handled in a consistent manner.

### Alternatives considered

_No response_

### Additional context

_No response_

### Submission Checklist

- [X] Once I hit submit, I will assign this issue to the Project Board with the appropriate tags.

### LINKED ISSUE #162: CRS should fetch only what is needed
### Problem

Currently, the backend fetches the entirety of transcript00, no less. This is a huge download for small circuits.

### Proposed solution

Instead of fetching all of transcript00, we should instead fetch only the amount of the G1 that is needed for a given circuit. 

### Alternatives considered

_No response_

### Additional context

This is being done in combination with #139

### Submission Checklist

- [X] Once I hit submit, I will assign this issue to the Project Board with the appropriate tags.

phated:
This is a draft implementation of https://github.com/noir-lang/acvm/pull/231 to see how the API worked out.

Closes #160
Closes #162


acvm_backend_barretenberg/src/composer.rs
```diff
@@ -142,15 +141,14 @@ impl Composer for Barretenberg {
...
-                &g2_data,
+                g2_data,
                 &proving_key,
                 &cs_buf,
                 &witness_buf,
                 p_proof,
             );
         }
 
-        std::mem::forget(g2_data);
```
vezenovm, line 153: Why can we remove this forget? Is g2_data not still being free'd inside bberg?

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -165,14 +163,15 @@ impl Composer for Barretenberg {
...
+        reference_string: &CRS,
         constraint_system: &ConstraintSystem,
         // XXX: Important: This assumes that the proof does not have the public inputs pre-pended to it
         // This is not the case, if you take the proof directly from Barretenberg
         proof: &[u8],
         public_inputs: Assignments,
         verification_key: &[u8],
     ) -> bool {
-        let g2_data = G2::new().data;
+        let CRS { g2_data, .. } = reference_string;
```
vezenovm, line 174: This change is good as it moves away from the G2 struct. One unified CRS is easier to reason about

acvm_backend_barretenberg/src/composer.rs
phated: I was just following clippy, but I think it is due to `g2_data` being of `&Vec<u8>` type so it's not needed 🤔

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -104,33 +104,32 @@ impl Composer for Barretenberg {
...
     }
 
     fn create_proof_with_pk(
         &self,
+        reference_string: &CRS,
         constraint_system: &ConstraintSystem,
         witness: WitnessAssignments,
         proving_key: &[u8],
     ) -> Vec<u8> {
-        let circuit_size = self.get_circuit_size(constraint_system);
```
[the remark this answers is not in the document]
phated: I didn't like the idea of persisting it on the struct because then you have to do like `Option<CRS>` and check to see if you have it in each function. The idea behind `get_reference_string` is that it won't be called by nargo if Nargo already has a cached version that is large enough for the circuit, so we'd have to check/cache it in each function, which seems messy.

acvm_backend_barretenberg/src/composer.rs
vezenovm: That may be the case as we are only fetching a reference to the Vec inside the method rather than it being shared amongst the composer? As long as we don't have seg faults this is fine with me, I just was surprised

phated: We removed all forget calls 🎉

src/crs.rs
```diff
@@ -1,146 +1,91 @@
...
-            download_crs(transcript_location()).unwrap();
-        }
-
-        // Read CRS, if it's incomplete, download it
-        let mut crs = read_crs(transcript_location());
-        if crs.len() < G2_END + 1 {
-            download_crs(transcript_location()).unwrap();
-            crs = read_crs(transcript_location());
-        }
+        // TODO(blaine): Make sure g1_start isn't off-by-one
```
TomAFrench, line 31: Could add a test where `download_crs(M)?.update(N) == download_crs(M+N)`? (For small values of M/N so that we're not downloading 300MB)

phated: Yeah, that's a good idea. We'll need to check it against a constraint system, but we can start small and then give a larger constraint system.

phated:
I'm see a lot of DNS failures on Linux. Do we know why or any way to avoid them? Is there a misconfiguration on the AWS bucket where the CRS lives? Or is it a problem with the Reqwest library (searching the internet for the error shows many similar errors from Reqwest...)?

We wouldn't have seen these errors with the old implementation because we were caching in CI and then we were using Nix to fetch.

cspell.json
```diff
@@ -55,6 +55,8 @@
         "envrc",
         "subshell",
         "thiserror",
+        "bincode",
+        "thiserror",
```
phated, line 59:
```suggestion
```
Duplicate

flake.nix
```diff
@@ -153,6 +149,27 @@
...
+          cp ${pkgs.barretenberg-transcript00} .
+          echo "Starting simple static server"
+          ${pkgs.simple-http-server}/bin/simple-http-server --port ${toString port} --silent &
+          HTTP_SERVER_PID=$!
+        '';
+
+        postCheck = ''
+          kill $HTTP_SERVER_PID
+        '';
+      };
```
phated, line 171: The network/dns failures on linux were due to the Nix sandbox disabling networking (it doesn't disable networking on mac for historical reasons).

So I needed to mimic our downloading locally in the sandbox, as seen in this code. Copying @kobyhallx for review, since it was his guidance in determining a solution.

src/composer.rs
```diff
@@ -1,30 +1,51 @@
...
     fn get_circuit_size(&self, constraint_system: &ConstraintSystem) -> Result<u32, Error>;
 
     fn get_exact_circuit_size(&self, constraint_system: &ConstraintSystem) -> Result<u32, Error>;
 
-    fn compute_proving_key(&self, constraint_system: &ConstraintSystem) -> Result<Vec<u8>, Error>;
-    fn compute_verification_key(
+    async fn get_crs(&self, constraint_system: &ConstraintSystem) -> Result<CRS, Error> {
+        let num_points = self.get_circuit_size(constraint_system)?;
+
+        download_crs(num_points as usize).await
```
kobyhallx, line 18: Should there be an option to read from a cache? Proposed in #184

flake.nix
kobyhallx: This looks good.

src/composer.rs
phated: I believe this conflicts with https://github.com/noir-lang/noir/issues/1302

PR #141, opened 2023-04-26
chore: inline contents of `common/proof.rs` into `acvm-backend-barretenberg`

TomAFrench:
These functions are only used in a single file each so I've moved them to the relevant locations.


vezenovm (approved):
I'm assuming these function starting being used in one place after the `barretenberg_static_lib` and `barretenberg_wasm` merge?

TomAFrench:
Yep, previously they were used in the `composer.rs` file for both of those packages but now those are both the same file.

PR #142, opened 2023-04-26
chore: miscellaneous refactoring of `barretenberg_structures`

TomAFrench:
I've removed some unused methods, derived default rather than defining it in terms of `new()` and replaced the usage of `Scalar` renaming of `FieldElement`.

### DIFF
--- a/acvm_backend_barretenberg/src/acvm_interop/smart_contract.rs
+++ b/acvm_backend_barretenberg/src/acvm_interop/smart_contract.rs
@@ -71,7 +71,7 @@ fn test_smart_contract() {
     use crate::composer::Composer;
     use crate::Barretenberg;
     use common::acvm::FieldElement;
-    use common::barretenberg_structures::ConstraintSystem;
+    use common::barretenberg_structures::{Constraint, ConstraintSystem};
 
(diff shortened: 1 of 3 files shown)


vezenovm (changes requested):
I like the separation of `Scalar` and `FieldElement` as we use field element everywhere as it is our base field. Scalar differentiates that this field element has to be used as part of the witness, not anywhere else

vezenovm:
> 've removed some unused methods, derived default rather than defining it in terms of `new()`

These switches I support merging

phated:
> I like the separation of Scalar and FieldElement as we use field element everywhere as it is our base field. Scalar differentiates that this field element has to be used as part of the witness, not anywhere else

Using `as` doesn't actually limit where you can use Scalar because Rust will accept a FieldElement. If you want this functionality, we need to introduce a newtype.

vezenovm (commented):
I will unblock my requested changes as this is just a renaming and making a new type is probably too restrictive. I just liked knowing that these FieldElements are to be used with Barretenberg when reading the code, but I don't feel super strongly about this

vezenovm (approved):

PR #143, opened 2023-04-26
chore: replace `check_membership` with `compute_merkle_root` 

TomAFrench:
Related to https://github.com/noir-lang/acvm/issues/66

I've also inlined the testing for merkle trees into `pwg.rs`. A lot of the merkle logic is unnecessary imo as it's related to constructing a merkle tree rather than just proving against an existing one. I'm then planning on removing this in a later PR.


acvm_backend_barretenberg/src/acvm_interop/pwg.rs
```diff
@@ -162,3 +165,279 @@ impl PartialWitnessGenerator for Barretenberg {
         Ok(OpcodeResolution::Solved)
     }
 }
+
+// TODO: alter this method so that it only processes one hash per level rather than overriding
+// the one of leaves for each level of the hash path
+fn calculate_merkle_root(
```
vezenovm, line 171: Why `calculate_merkle_root`? Let's keep it `comptue_merkle_root` like the opcode?

acvm_backend_barretenberg/src/acvm_interop/pwg.rs
```diff
@@ -162,3 +165,279 @@ impl PartialWitnessGenerator for Barretenberg {
         Ok(OpcodeResolution::Solved)
     }
 }
+
+// TODO: alter this method so that it only processes one hash per level rather than overriding
+// the one of leaves for each level of the hash path
```
TomAFrench, line 170:
```suggestion
```

This comment looks stale to me as it looks like we're doing the minimal number of hashes.

acvm_backend_barretenberg/src/acvm_interop/pwg.rs
```diff
@@ -162,3 +165,279 @@ impl PartialWitnessGenerator for Barretenberg {
...
+            } else {
+                (path_elem, &current_node)
+            };
+            hash_func(left, right)
+        },
+    )
+}
+
+#[cfg(test)]
+mod tests {
```
vezenovm, line 198: I don't see the need to bring these tests and merkle root calculation into `pwg.rs`. I think pwg.rs should only contain the PartialWitnessGenerator. If we want to remove the merkle.rs file from the `pwg` folder is there a way we could have a `Merkle` trait that impls on Barretenberg similarly to Pedersen and Schnorr (although this may be more confusing as we are not using barretenberg for our merkle logic, only the hash function).

acvm_backend_barretenberg/src/acvm_interop/pwg.rs
vezenovm: Yes it is stale

vezenovm:
> A lot of the merkle logic is unnecessary imo as it's related to constructing a merkle tree rather than just proving against an existing one. I'm then planning on removing this in a later PR.

Yeah it is only related to constructing a merkle tree and is really only useful for tests. We should be careful about removing things that we may have to potentially bring back. If we ever implement any sparse merkle tree functions or natively enable merkle trees of different arity we will want to expand upon these merkle tree classes

acvm_backend_barretenberg/src/acvm_interop/pwg.rs
TomAFrench: This is to match the proposed change to this opcode in https://github.com/noir-lang/acvm/issues/66.

vezenovm: Yes ik :) this is a nit, is prefer `compute_merkle_root` like we have it here (https://github.com/noir-lang/acvm/pull/233) so that the names match

TomAFrench: Ohhh, sorry I misread your comment. Agreed we should match the opcode.

TomAFrench:
> We should be careful about removing things that we may have to potentially bring back.

I'm of the opinion that that's best left to git to handle. If we need a merkle tree implementation then we can fish it back from the git history, otherwise we need to do all the work to maintain this code all while never using it.

These tests don't feel particularly useful to me as the vast majority of the complexity is actually in the testing code (the merkle tree implementation to construct the tree rather than the function which is verifying it).

acvm_backend_barretenberg/src/acvm_interop/pwg.rs
TomAFrench: I've moved this back up into `pwg/merkle.rs` .

vezenovm (approved):

vezenovm:
> I'm of the opinion that that's best left to git to handle. If we need a merkle tree implementation then we can fish it back from the git history, otherwise we need to do all the work to maintain this code all while never using it.

The dead code can be removed in a separate PR

PR #144, opened 2023-04-26
chore: pull CRS from `common` into `acvm-backend-barretenberg`

TomAFrench:
This pulls across the CRS downloading logic into `acvm-backend-barretenberg`.


acvm_backend_barretenberg/Cargo.toml
```diff
@@ -28,15 +35,18 @@ tempfile = "*"
 
 [features]
 default = ["native"]
+
+std = ["dep:reqwest", "dep:tokio", "dep:futures-util", "dep:dirs", "dep:indicatif"]
```
phated, line 39: I think we just drop this feature. These impls are going to be behind native/wasm/js feature flags. No reason to keep this.

acvm_backend_barretenberg/src/composer.rs
```diff
@@ -1,6 +1,6 @@
 use common::barretenberg_structures::*;
-use common::crs::{CRS, G2};
 
+use crate::crs::{CRS, G2};
```
phated, line 3: G2 is going away with reference string changes

phated (changes requested):

acvm_backend_barretenberg/src/composer.rs
TomAFrench: I think let's keep it for now. If changes are required for the new interface changes then we can adjust this at the same time.

acvm_backend_barretenberg/Cargo.toml
TomAFrench: Would you prefer to inline all of these dependencies into `native` and `wasm`? That seems noisy to me and allows for  them to get out of sync.

phated: Yeah, I think we should. I'd rather do work to keep the features in sync than to have confusing combinations of features.

acvm_backend_barretenberg/Cargo.toml
```diff
@@ -10,7 +10,14 @@ rust-version.workspace = true
 
 [dependencies]
 blake2 = "0.9.1"
-common = { path = "../common", default-features = false }
+common = { path = "../common" }
+
+# std
```
phated, line 15:
```suggestion
```
nit

acvm_backend_barretenberg/Cargo.toml
```diff
@@ -28,20 +35,29 @@ tempfile = "*"
 
 [features]
 default = ["native"]
+
```
phated, line 38:
```suggestion
```
nit

phated (approved):
lgtm, can you cleanup the "behaviour" cspell? Silly american english

PR #147, opened 2023-04-26
feat!: remove all crates other than `acvm-backend-barretenberg` and remove workspace

TomAFrench:


Cargo.toml
```diff
@@ -1,9 +1,65 @@
-[workspace]
-
-members = ["common", "acvm_backend_barretenberg", "aztec_backend_wasm"]
-
-[workspace.package]
+[package]
+name = "acvm-backend-barretenberg"
+version = "0.1.0"
```
phated, line 3:
```suggestion
version = "0.0.0"
```
So we can make the first release with RP and it'll be 0.1.0

phated (approved):
Just want to drop that version to 0.0.0, but otherwise looks good.

phated:
Ah, and the wasm workflow needs to be deleted

TomAFrench:
I'm also getting some issues in flake locally. I'm just checking if CI experiences the same atm.

TomAFrench:
Removing the todo as we should rename the repo after this PR.

phated:
I just removed the build wasm required check to get this queued

PR #148, opened 2023-04-26
feat!: migrate to ACVM 0.10.3

TomAFrench:
Blocked by https://github.com/AztecProtocol/barretenberg/pull/385 and https://github.com/noir-lang/acvm/pull/204

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,9 @@ license = "MIT OR Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-acvm = { version = "0.9.0", features = ["bn254"] }
+#acvm = { version = "0.9.0", features = ["bn254"] }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -9,7 +9,9 @@ license = "MIT OR Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-acvm = { version = "0.9.0", features = ["bn254"] }
```
kevaundray, line 12: Note: can be 0.10.1 now

src/composer.rs
```diff
@@ -761,6 +765,57 @@ mod test {
         test_composer_with_pk_vk(constraint_system, vec![case_1]);
     }
 
+    #[test]
+    fn test_compute_merkle_root_constraint() {
+        use tempfile::tempdir;
+        let temp_dir = tempdir().unwrap();
```
kevaundray, line 771: not blocking: perhaps we could find a better way to test this such that we can re-use infrastructure from the merkle tests

kevaundray (approved):

PR #150, opened 2023-04-28
feat!: Add Keccak constraints

guipublic:
This PR includes Keccak constraints when it serialise an ACIR circuit having Keccak opcodes.
It must be used with a Barretenberg version which handle them (cf. https://github.com/AztecProtocol/barretenberg/pull/393)


src/barretenberg_structures.rs
```diff
@@ -302,6 +302,34 @@ impl HashToFieldConstraint {
...
     }
 }
+
+#[derive(Clone, Hash, Debug)]
+pub(crate) struct HashConstraint {
+    pub(crate) inputs: Vec<(i32, i32)>,
+    pub(crate) result: [i32; 32],
+}
+
+impl HashConstraint {
```
TomAFrench, line 312: This name is too generic as we have many hash constraints

TomAFrench (changes requested):
This doesn't currently implement support for keccak in the `PartialWitnessGenerator`

src/barretenberg_structures.rs
guipublic: Yes, it is on purpose, I plan to use the same structure for sha256 and blake2s.

TomAFrench: We can rename it when this struct has been generalised in that case.

kevaundray:
@TomAFrench I have renamed the `HashConstraint` struct to be `Keccak256Constraint`.

Since the pwg function for Keccak256 was put in after the fallible traits, PR 151 would be a blocker for the Keccak256 opcode unless we just implemented the Keccak256 functionality in aztec_backend until we can land acvm >0.11. I left a comment [here](https://github.com/noir-lang/aztec_backend/pull/151#issuecomment-1535472605) on 151 and spoke to @phated on slack. I went with re-implementing it in here so we don't rush 151 and have it come out in a less than ideal state, coupled with the fact that the change is isolated and mild.

Let me know your thoughts on this

TomAFrench:
Looks good but I think we expose keccak256 from ACVM now so we can remove the vendored code. Will do that now.

TomAFrench (approved):
Tested this locally and it works so should work in CI

TomAFrench:
We should add a simple testcase for a keccak constraint. I'm ok with delaying this as we want to get this out for nouns though.

PR #153, opened 2023-05-03
feat!: serialise RAM and ROM opcodes

guipublic:
This PR serialise RAM and ROM opcodes from ACIR into BB acir format.


TomAFrench:
Can we add a couple of test cases similar to [this](https://github.com/noir-lang/aztec_backend/blob/c9fb9e806f1400a2ff7594a0669bec56025220bb/src/composer.rs#L716)? Ideally we'd have an test for each constraint type so we don't rely on noir integration tests to check we're handling them correctly here.

guipublic:
I added a unit test, it is failing because it cannot locate barretenberg so I cannot test it, but anyways the relevant barretenberg PR is not merge yet.

guipublic:
> Huh, looks like the upstream BB PR hasn't been merged yet. Strange that this PR's test is passing in that case.

I suppose it is because I added the new data at the end when serialising, so if you read less, you still get valid data.

TomAFrench:
I've bumped the BBerg version to include the relevant PRs so the test should be actually testing stuff now.

TomAFrench:
It would be good to see a commit which breaks the blocks test case (e.g. use a witness which doesn't satisfy the block constraints) to show that this test is meaningful and then revert it.

vezenovm:
> It would be good to see a commit which breaks the blocks test case (e.g. use a witness which doesn't satisfy the block constraints) to show that this test is meaningful and then revert it.

Or just leave it in and show verify false

src/barretenberg_structures.rs
```diff
@@ -641,13 +648,91 @@ impl ConstraintSystem {
...
+
+        buffer
+    }
+}
+
+#[derive(Clone, Hash, Debug)]
+pub(crate) struct MemOpBarretenberg {
+    pub(crate) index: Constraint,
+    pub(crate) value: Constraint,
+    pub(crate) is_store: i8,
```
vezenovm, line 666: Let's make this `is_store` the first value so it matches with the struct in barretenberg and it is clear why we are writing it first in `to_bytes`

vezenovm:
Just a small nit

guipublic:
> > It would be good to see a commit which breaks the blocks test case (e.g. use a witness which doesn't satisfy the block constraints) to show that this test is meaningful and then revert it.
> 
> Or just leave it in and show verify false

I have added a failing test

src/barretenberg_structures.rs
guipublic: done

src/barretenberg_structures.rs
```diff
@@ -641,13 +648,91 @@ impl ConstraintSystem {
...
+
+impl BlockConstraint {
+    fn to_bytes(&self) -> Vec<u8> {
+        let mut buffer = Vec::new();
+
+        let len = self.init.len() as u32;
+        buffer.extend_from_slice(&len.to_be_bytes());
+        for value in self.init.iter() {
+            buffer.extend_from_slice(&value.to_bytes());
+        }
```
TomAFrench, line 695: Not a blocker to this PR but we should factor this logic out as it's used so much.

TomAFrench (approved):

TomAFrench (approved):

vezenovm (approved):

PR #154, opened 2023-05-04
chore!: replace `black_box_function_supported` with `supports_opcode`

guipublic:
This PR allows the backend to indicate which opcode they support.
For instance we could have a TP backend which supports ROM opcode and not the RAM opcode.


TomAFrench:
For context on this, we'll also need to modify the `isOpcodeSupported` type so that this will be usable in `acvm::compile` due to one of the issues in https://github.com/noir-lang/acvm/issues/183

These changes need to be made in ACVM so does't affect this PR directly.

TomAFrench:
This needs to have it's base changed to the ACVM 0.12.0 branch.

PR #164, opened 2023-05-08
chore: remove usage of `std::mem::forget`

TomAFrench:
It's not clear to me on why we "forget" the CRS, PK, constraint systems and witness. I remember that previously Barretenberg would have some singleton instance of the proving key, etc. which would make some sense. We don't do this anymore and in any case, we pass in new values for each of these on every call so I don't see any reason to not to free these resources.

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -74,8 +74,6 @@ impl Composer for Barretenberg {
             pk_size = barretenberg_sys::composer::init_proving_key(&cs_buf, pk_ptr);
         }
 
-        std::mem::forget(cs_buf);
-
         let result;
         unsafe {
             result = Vec::from_raw_parts(pk_addr, pk_size, pk_size);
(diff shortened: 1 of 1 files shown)


vezenovm (dismissed):
We forget these pointers because their destructor will be called by the C++. We don't want to free memory that has already been freed. It is not that ownership of these values are shared amongst Rust, but rather than ownership is transferred to the C++. I would rather move to something like `ManuallyDrop` vs. getting rid of the forget as it can lead to some funky bugs. 

I see our `aztec_backend` tests are passing on this, but I would like this to be integration tested with Noir before these are removed. 

@kevaundray does something come to mind about what may be good to test removing these `forget` calls?

phated (approved):
As per @kevaundray's comment and the keccak test suite passing on this CI run: https://github.com/noir-lang/noir/actions/runs/4920482191/jobs/8789322898?pr=1249

I also did a memory profile and found the pkey and vkey were not being freed.
Screenshot 2023-05-08 at 3 55 18 PM

TomAFrench:
Well this was good timing :sweat_smile:

PR #165, opened 2023-05-08
feat!: Update to acvm 0.12.0

TomAFrench:
Updates the ACVM dependency to current master of ACVM.

This has shown an unfortunate effect of https://github.com/noir-lang/acvm/pull/257 as we need to reconstruct the `BlackBoxFuncCall` which gets deconstructed in ACVM in order to call back into the solver functions.


src/acvm_interop/pwg.rs
```diff
@@ -11,185 +12,297 @@ use crate::scalar_mul::ScalarMul;
...
-                let hash_path: Result<Vec<_>, _> = inputs_iter
-                    .map(|input| witness_to_value(initial_witness, input.witness))
-                    .collect();
-
-                let computed_merkle_root = merkle::compute_merkle_root(
-                    |left, right| self.compress_native(left, right),
-                    hash_path?,
-                    index,
-                    leaf,
+        logic::and(
```
phated, line 35: I think the ACVM functions need to be changed. It seems weird to be building this struct to call them.

TomAFrench: Yep, I made https://github.com/noir-lang/acvm/pull/268 to address this.

phated:
I rebased this because I need it as a basis on my CRS changes that'll land in 0.12

TomAFrench:
Yep, I'm happy for anyone to update this to target the ACVM master as needed.

kevaundray:
This can now be updated to target 0.12 on crates.io

phated:
A lot of folks have touched this, so I'm going to request reviews from folks that haven't (such as @kevaundray @kobyhallx and @vezenovm)

src/acvm_interop/proof_system.rs
```diff
@@ -21,58 +19,77 @@ impl ProofSystemCompiler for Barretenberg {
...
-            | BlackBoxFunc::SchnorrVerify
-            | BlackBoxFunc::Pedersen
-            | BlackBoxFunc::HashToField128Security
-            | BlackBoxFunc::EcdsaSecp256k1
-            | BlackBoxFunc::FixedBaseScalarMul => true,
-
-            BlackBoxFunc::AES => false,
+            Opcode::Arithmetic(_) => true,
+            Opcode::Directive(_) => true,
+            Opcode::Block(_) => false,
```
vezenovm, line 26: We don't support `Block`? cc'ing @guipublic

src/crs.rs
```diff
@@ -1,194 +1,172 @@
...
 }
 
 impl CRS {
-    pub(crate) fn new(num_points: usize) -> CRS {
+    pub(crate) async fn update(&mut self, num_points: usize) -> Result<(), Error> {
+        // We already have some data, so start at the end of our list
+        let g1_start = G1_START + self.g1_data.len();
         // UltraPlonk requires a CRS equal to circuit size plus one!
         // We need to bump our polynomial degrees by 1 to handle zero knowledge
         let g1_end = G1_START + ((num_points + 1) * 64) - 1;
```
vezenovm, line 31: Should we add a sanity check that g1_and is greater than g1_start? Just in case `((num_points + 1) * 64) - 1` is smaller than `self.g1_data.len()`?

vezenovm (commented):
A couple minor comments. Not really familiar with the nix stuff so I defer to @phated and @kobyhallx on that

src/acvm_interop/common_reference_string.rs
```diff
@@ -0,0 +1,34 @@
+use acvm::{acir::circuit::Circuit, async_trait, CommonReferenceString};
+
+use crate::{composer::Composer, BackendError, Barretenberg};
+
+// TODO: Separate impl for JS feature
```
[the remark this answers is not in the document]
phated:
```suggestion
// TODO(#185): Ensure CRS download works in JS
```
I'm unifying them and I tested it browser and it works; just need to upstream a bunch of needed changes from my WIP

src/crs.rs
phated: We actually do the num_points check in the composer API, but I'm fine wrapping this in an `if`. Pushing now.

vezenovm: Ah I missed that you do the `num_points` check there. Perhaps just note it in the method then? It seems unlikely that anyone else would ever be consuming the CRS API

src/acvm_interop/proof_system.rs
kevaundray: It doesn't matter that much right now since the current Noir code will just use the block opcode anyways.

It will matter more depending on how 1282 is implemented. Block is a bit weird because its sort of like the fallback implementation for ROM/RAM

TomAFrench: Sounds like this PR is good to go then?

kevaundray: Yeah if that was the only blocker, then we can tenatively merge

kevaundray (approved):

PR #168, opened 2023-05-09
chore: Add sanity checks for the number of bytes in `FieldElement`

TomAFrench:
This PR adds some sanity check tests for the field which we're compiling against. i.e. the number of bytes to represent a `FieldElement` is 32 (to match what bberg expects) and so we can call `field_to_array` without it panicking.


src/lib.rs
```diff
@@ -118,8 +118,36 @@ impl From<CRSError> for BackendError {
...
+// i.e. it can be represented in `FIELD_BYTES` bytes and can be converted to a byte array of that length.
+#[cfg(test)]
+mod field_size_checks {
+    use acvm::FieldElement;
+
+    use crate::FIELD_BYTES;
+
+    #[test]
+    fn check_field_size() {
+        assert_eq!(FIELD_BYTES, FieldElement::max_num_bytes() as usize);
```
phated, line 136: Is there a reason we aren't using `FieldElement::max_num_bytes()` within our code instead of setting the FIELD_BYTES constant?

src/lib.rs
```diff
@@ -146,20 +174,23 @@ fn smoke() -> Result<(), Error> {
 
 #[cfg(feature = "native")]
 mod native {
-    use super::{Barretenberg, Error, FeatureError};
+    use crate::FIELD_BYTES;
+
+    use super::{Barretenberg, FeatureError};
```
phated, line 179:
```suggestion
    use super::{Barretenberg, FeatureError, FIELD_BYTES};
```

src/lib.rs
TomAFrench: A couple of reasons:
- `FieldElement::max_num_bytes()` is much more verbose than `FIELD_BYTES`
- `FieldElement::max_num_bytes()` is a `u32` whereas a lot of the usage of `FIELD_BYTES` requires a `usize` so we need to do casting in function signatures which is yucky.

PR #170, opened 2023-05-09
chore: add `SchnorrSignature` struct

TomAFrench:
This helps avoid us having to merge and split the signature element arrays a couple of times, avoiding the situation where `SchnorrConvert` is thrown in #151. It also motivates removal of the `SchnorrSlice`  error variant .

### DIFF
--- a/src/acvm_interop/pwg.rs
+++ b/src/acvm_interop/pwg.rs
@@ -100,7 +100,7 @@ impl PartialWitnessGenerator for Barretenberg {
                     message.push(msg_i);
                 }
 
-                let valid_signature = self.verify_signature(pub_key, signature, &message);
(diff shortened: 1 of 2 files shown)


phated:
These errors stemmed from us trying to construct a "signature" when we always needed a `sig_s` and `sig_e`. Instead of trying to construct & deconstruct all over the backend, I've instead just using `sig_s` and `sig_e` everywhere. The only place this had significant impact was in the Circuit -> ConstraintSystem conversion, but that uses iterators so it was fine too.

Superseded by https://github.com/noir-lang/aztec_backend/pull/151/commits/8c56b112b9f045f2ee7b93d88ef7eb8c809cc5f9 in #151

PR #190, opened 2023-05-16
chore: Remove `sled` & `tempfile` dev-dependencies

phated:
This removes `sled` and `tempfile` from the merkle tests. By removing these file-based dependencies from the test infrastructure, we'll need less separation between between wasm32 target tests and native tests.

I'm not sure why the file-based dependencies were used, but it seems that the tests can be done with in-memory BTreeMaps.

Tests are passing, but need a review to make sure I didn't mess something up which makes them always pass 😛

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,8 +61,6 @@ dependencies = [
  "pkg-config",
  "reqwest",
  "rust-embed",
- "sled",
- "tempfile",
  "thiserror",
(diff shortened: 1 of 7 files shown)


TomAFrench (approved):
LGTM, we can make some more simplifications here imo but low priority for now.

e.g. self.db` only holds 3 named values so we can just store those in separate fields.

PR #191, opened 2023-05-16
chore: Switch to tokio test macro for async function

phated:
This switches to using the `tokio::test` macro to use tokio async test runner for any functions that rely on CRS and other async functions in the future.

Removing the tokio calls from functions will allow us to swap in other test runners that support async functions, such as wasm-bindgen-test for browser tests.

I also made the CRS pippenger test run in all targets by relying on the `get_pippenger` abstraction.

### DIFF
--- a/src/acvm_interop/smart_contract.rs
+++ b/src/acvm_interop/smart_contract.rs
@@ -75,47 +75,50 @@ impl SmartContract for Barretenberg {
     }
 }
 
-#[test]
-fn test_smart_contract() -> Result<(), BackendError> {
-    use crate::barretenberg_structures::{Constraint, ConstraintSystem};
-    use crate::composer::Composer;
-    use crate::Barretenberg;
-    use acvm::FieldElement;
-    use tokio::runtime::Builder;
+#[cfg(test)]
+mod tests {
+    use acvm::SmartContract;
+    use tokio::test;
 
-    let constraint = Constraint {
-        a: 1,
-        b: 2,
-        c: 3,
-        qm: FieldElement::zero(),
-        ql: FieldElement::one(),
-        qr: FieldElement::one(),
-        qo: -FieldElement::one(),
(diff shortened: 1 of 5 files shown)


src/acvm_interop/smart_contract.rs
```diff
@@ -75,47 +75,50 @@ impl SmartContract for Barretenberg {
...
-
-    Ok(())
+#[cfg(test)]
+mod tests {
+    use acvm::SmartContract;
+    use tokio::test;
+
+    use crate::BackendError;
+
+    #[test]
```
TomAFrench, line 85:
```suggestion
    use acvm::SmartContract;

    use crate::BackendError;

    #[tokio::test]
```

Not going to block on this but I prefer the qualified usage of this to make is clearer that we're running this test inside tokio. I can follow up with this change in a separate PR however if you don't have strong feelings on why you did it this way.

TomAFrench (approved):

TomAFrench:
tbh, I'm inclined to rework/remove the `does_not_panic` test as it's a bit muddled on what it's trying to test. If it's testing the CRS then why are we creating a pippenger pointer? There doesn't seem to be any validation in bberg on the CRS when creating a pippenger object so it's not checking for garbage data.

`//TODO check that p_points memory is properly free` seems to be purely a pippenger concern.

src/acvm_interop/smart_contract.rs
phated: I purposefully imported it that way, because I will use conditional compilation to `use wasm_bindgen_test::wasm_bindgen_test as test` and then the macros on each don't need to be duplicated nor do we need to create separate tests.

TomAFrench: Gotcha, yeah I noticed this would make it easier for conditional compilation which was the main reason I didn't go ahead with the change.

PR #197, opened 2023-05-17
chore: Replace wasmer uint8view with view calls

phated:
This replaces the usage of wasmer's `memory.uint8view()` function and replaces it with `memory.view()`. It seems that the return type of `uint8view` differs by platform, but `view` is consistent.

My experiments indicate that these implementations will compile on native or wasm32 targets, but I'm leaving it a draft until that can be proven by an actual wasm32 build.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,7 +138,6 @@ mod native {
 
 #[cfg(not(feature = "native"))]
 mod wasm {
-    use std::cell::Cell;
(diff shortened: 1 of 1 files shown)


phated:
This is actually flagged as `unimplemented!()` in wasmer 2.3 but it is implemented in wasmer 3, which we are going to be switching to.

TomAFrench:
Seems like this is superceded by #208?

PR #198, opened 2023-05-17
fix: Allow async functions without send on async trait

phated:
As noted in https://github.com/noir-lang/acvm/pull/292, the JS implementation does not implement the `Send` trait, so this allows our async code to not implement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,7 +5,7 @@ version = 3
 [[package]]
 name = "acir"
 version = "0.12.0"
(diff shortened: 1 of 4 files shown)


phated:
I changed the base to the acvm 0.13 branch since that's where this change would land.

TomAFrench:
Happy to merge as soon as ACVM commit has been updated.

TomAFrench (approved):

PR #199, opened 2023-05-18
fix: Add or cleanup implementations for JS target

phated:
This is a temporary fix until we can get wasmer 3.x and then we can join the implementations like #197

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,7 +138,6 @@ mod native {
 
 #[cfg(not(feature = "native"))]
 mod wasm {
-    use std::cell::Cell;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -405,14 +406,29 @@ mod wasm {
...
+            if cell.get() != 0 {
                 ptr_end = i;
             } else {
                 break;
             }
         }
 
+        #[cfg(feature = "js")]
+        let str_vec: Vec<_> =
+            byte_view.to_vec()[ptr as usize..=(ptr + ptr_end as i32) as usize].to_vec();
```
TomAFrench, line 429: Can we not use `.subarray()` here?

PR #202, opened 2023-05-18
feat!: Replace `js` feature with `wasm32` target

phated:
Depends on #198 and #199 

This reorganizes the `Cargo.toml` to tie features to their dependencies and just enables the dependency based on the `native` or `wasm` feature of the crate. It also removes the `js` feature and instead relies on the wasm32 `target_arch`.


TomAFrench (approved):
LGTM

TomAFrench:
Actually, on second thoughts... As people can't currently build for JS with the `js` flag then surely this is non-breaking?

PR #203, opened 2023-05-19
chore: add wasm fields to `Barretenberg` with `js` flag

TomAFrench:
We currently only add these fields with the `wasm` feature flag but we also need it for the `js` flag. We then always include these if we're not using the native barretenberg.


TomAFrench:
superceded by #202

PR #204, opened 2023-05-19
feat: allow updating an empty CRS buffer

TomAFrench:
Resolves https://github.com/noir-lang/noir/pull/1348#discussion_r1198654149


phated (changes requested):
From a caller/callee perspective, I think it's wrong to ever call the update function with an empty vector. The reason that an empty vector is used in Nargo is to make the type system happy; otherwise we'd be dealing with options in a lot of places I didn't want to.

There may be a case for only having a `get_or_update_crs` API but that'd take an Option for CRS and the `is_empty` check would still need to exist in Nargo.

PR #214, opened 2023-05-31
feat!: Add latest BB to acvm 0.13

sirasistant:


src/pedersen.rs
```diff
@@ -117,8 +117,8 @@ impl Pedersen for Barretenberg {
         let result_ptr: usize = 0;
 
         self.call_multiple(
-            "pedersen_plookup_commit",
-            vec![&input_ptr, &result_ptr.into()],
+            "pedersen_plookup_commit_with_hash_index",
+            vec![&input_ptr, &result_ptr.into(), &hash_index.into()],
```
vezenovm, line 121: hash_index is never passed into the encrypt method thus you are getting an error that `hash_index` is not found in this scope

sirasistant: Yeah we need the updated ACVM to have the value, I cherrypicked from the complete changes and forgot to fix that. Will fix now, thx!

PR #219, opened 2023-06-20
fix: Remove wasm32 target

phated:
# Description

## Problem\*

Resolves 

## Summary\*

It seems that `noir_wasm` causes dependency conflicts within this crate when we enable the wasm backend. Removing the wasm32 target from this package allows us to build with the wasm backend again.

Since we aren't going to use this package in the browser, we can just remove this instead of a larger refactoring.

## Additional Context

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - uses: cachix/install-nix-action@v20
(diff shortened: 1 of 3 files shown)


kobyhallx (approved):
LGTM

PR #239, opened 2023-08-17
feat!: Call backend via precompiled binaries

### LINKED ISSUE #242: Replace static library/wasm binary with a bundled barretenberg binary
This is a stepping stone to full dynamic backends where we can handle the issues around downloading a backend and communicating with it through the CLI before we loosen the coupling.

kevaundray:
# Description

## Problem\*

Resolves 

## Additional Context

# PR Checklist\*


kevaundray:
Changes made based on: https://github.com/AztecProtocol/aztec-packages/pull/1672

build.rs
```diff
@@ -7,7 +11,27 @@ use std::env;
 //     }
 // }
 
+const USERNAME: &str = "AztecProtocol";
+const REPO: &str = "barretenberg";
+const VERSION: &str = "0.3.6";
```
kevaundray, line 16:
```suggestion
const VERSION: &str = "0.4.2";
```

src/bb/contract.rs
```diff
@@ -26,7 +25,7 @@ impl ContractCommand {
             .arg("-k")
             .arg(self.path_to_vk)
             .arg("-o")
-            .arg(self.path_to_contract_output);
+            .arg("/dev/null");
```
kevaundray, line 28: Note: windows does not have /dev/null !

TomAFrench: Windows has `NUL` however which we can use here.

kevaundray:
Hmm I don't think Mac builds should be taking 1 hour plus

kevaundray:
This should still fail on mac since that binary is not patched.

I found two problems:

- The bb binary does not come with libstdc++ , so we had to point it to where it is in the nix sandbox. You can check by doing ldd ./backend_binary  . 

This was fixed by patching the binary with `patchelf --set-rpath "${stdenv.cc.cc.lib}/lib:${pkgs.gcc.cc.lib}/lib" ./backend_binary`

- It was pointing to the wrong interpretter. You can check this by running `readelf -l ./backend_binary | grep "interpreter"` It should say /lib64/ld-linux-x86-64.so.2. 

This is what it is for normal linux systems, but if you are in the sandbox it is actually something like `/nix/store/...ld-linux-x86-64.so.2` . If you run `ls /lib64/ld-linux-x86-64.so.2` it will come up with missing file, which is likely the error that we were encountering because the interpretter is called when the binary is called. 

This was fixed by patching it with `patchelf --set-interpreter ${stdenv.cc.libc}/lib/ld-linux-x86-64.so.2 ./backend_binary`

kevaundray:
Seems the Mac binary doesn't need patching

TomAFrench (dismissed):

TomAFrench:
Merging with agreement from @kevaundray

.github/workflows/test.yml
```diff
@@ -11,14 +11,15 @@ jobs:
...
+    timeout-minutes: 30
 
     strategy:
       fail-fast: false
       matrix:
         include:
           - os: ubuntu-latest
             target: x86_64-linux
-          - os: macos-latest
+          - os: macos-latest-xl
```
kevaundray, line 22:
```suggestion
          - os: macos-latest
```

kevaundray:
Checking build times once again in https://github.com/noir-lang/noir/pull/2434 -- Lets merge in #251 , merge this in and then followup by pushing the wasm stuff into acvm

src/bb/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+fn get_bb_download_url() -> String {
+    if let Ok(path) = std::env::var("BB_BINARY_URL") {
+        return path;
+    }
+
+    let target_os = env!("TARGET_OS");
+    let target_arch = env!("TARGET_ARCH");
+
+    let archive_name = match target_os {
+        "linux" => "bb-ubuntu.tar.gz",
```
kevaundray, line 46:
```suggestion
        "linux" => "barretenberg-x86_64-linux-gnu.tar.gz",
```

src/bb/gates.rs
```diff
@@ -0,0 +1,71 @@
...
+                String::from_utf8_lossy(&output.stderr)
+            );
+        }
+        // Note: barretenberg includes the newline, so that subsequent prints to stdout
+        // are not on the same line as the gates output.
+
+        // Ensure we got the expected number of bytes
+        if output.stdout.len() != 8 {
+            panic!("Unexpected 8 bytes, received {}", output.stdout.len());
+        }
```
TomAFrench, line 35: This panics on my machine.

TomAFrench (approved):

